PR #73, opened 2014-12-31
Allow setting CDN, API URL and HTTP proxy settings in Hex config

JamesS237:
This pull request allows users to set the following configuration settings from the config file:

- API URL
- CDN URL
- HTTP proxy
- HTTPS proxy

>Note: Sadly, setting the home directory from the config file is impossible.

This allows users to do the following:
```shell
$ mix hex.config api_url http://hex.internal.corporation.com
$ mix hex.config cdn_url http://cdn.hex.internal.corporation.com
```

:christmas_tree: :fireworks: Merry Christmas and a Happy New Year!

### DIFF
--- a/lib/hex.ex
+++ b/lib/hex.ex
@@ -3,7 +3,6 @@ defmodule Hex do
   @default_cdn "https://s3.amazonaws.com/s3.hex.pm"
   @default_home "~/.hex"
 
-
   def start do
     unless Application.get_env(:hex, :started) do
       start_api()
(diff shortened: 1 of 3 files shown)


JamesS237:
screen shot 2015-01-01 at 4 49 55 pm
Here's the new docs rendered.

josevalim:
Beautiful!

lib/hex.ex
```diff
@@ -18,12 +17,24 @@ defmodule Hex do
...
+    if url  = Keyword.get(config, :api_url), do: url(url)
+    if cdn  = Keyword.get(config, :cdn_url), do: cdn(cdn)
+
+    # HTTP proxy configuration from both environment variables and Hex config
+    http_proxy  = System.get_env("HTTP_PROXY")
+                  || System.get_env("http_proxy")
+                  || Keyword.get(config, :http_proxy)
+    https_proxy = System.get_env("HTTPS_PROXY")
+                  || System.get_env("https_proxy")
+                  || Keyword.get(config, :https_proxy)
```
ericmj: Env vars should override the config, not the other way override.

lib/mix/tasks/hex/config.ex
```diff
@@ -7,6 +7,16 @@ defmodule Mix.Tasks.Hex.Config do
...
+
+  ## Default:
+   - `username` - Hex username
+   - `key` - Hex API key
+
+  ## Optional:
+   - `api_url` - Hex API base URL (without trailing slash)
+   - `cdn_url` - Hex CDN base URL (without trailing slash)
+   - `http_proxy` - HTTP proxy server
+   - `https_proxy` - HTTPS proxy server
```
ericmj: No colons in headers and I think using the word default is wrong here. What do you think about this?

```
 ## Config keys
  - `username` - Hex username
  - `key` - Hex API key
  - `api_url` - Hex API base URL (without trailing slash)
  - `cdn_url` - Hex CDN base URL (without trailing slash)
  - `http_proxy` - HTTP proxy server
  - `https_proxy` - HTTPS proxy server

ericmj: We should trim the trailing slash instead of documenting that it isn't allowed.

JamesS237: I agree on both of these points here. Fixing up and rebasing.

lib/hex.ex
JamesS237: Agreed.

lib/mix/tasks/hex/config.ex
JamesS237: Turns out we already trim the trailing slash! Whoops

JamesS237:
All fixed and rebased. Ready to go!

ericmj:
Thanks James!

PR #75, opened 2015-01-02
Prefer doc/ for documentation.

### LINKED ISSUE #72: hexdocs should give preferences to the doc directory
Today, the hex.docs task checks for the docs directory. We should instead check for the doc directory (singular) and, if not available, check for the docs directory. The reason is that OTP uses doc and we should use it too. Basically, the steps are:

1. Check doc exists, if so, use it
2. Check docs exists, if so, print a warning "Your documentation was generated in the docs directory. If you are using ex_doc, please update to the latest which will generate the documentation correctly at the doc directory"
3. If none is available, fail with docs not available

I will work on a fix for ExDoc and we can push new versions around the same time. :heart:

/cc @JamesS237 @ericmj

JamesS237:
Closes #72 

Changes:
  - allows for both `doc/` and `docs/` documentation directories
  - raises an error when documentation directory is unavailable
  - alerts ex_doc users that they can update to the latest release that will prefer `doc/`

/cc @josevalim 

:page_with_curl:

### DIFF
--- a/lib/mix/tasks/hex/docs.ex
+++ b/lib/mix/tasks/hex/docs.ex
@@ -40,8 +40,13 @@ defmodule Mix.Tasks.Hex.Docs do
     else
       Mix.Task.run("docs", args)
 
-      unless File.exists?("docs/index.html") do
-        Mix.raise "File not found: docs/index.html"
+      if File.exists?("docs/") do
+          Mix.shell.info([IO.ANSI.yellow, "Your documentation was generated in the docs/ directory.", IO.ANSI.reset])
+          Mix.shell.info([IO.ANSI.yellow, "If you are using ex_doc, please update to the latest release which will generate the documentation correctly at the doc/ directory.", IO.ANSI.reset])
+      end
+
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/hex/docs.ex
```diff
@@ -40,8 +40,11 @@ defmodule Mix.Tasks.Hex.Docs do
     else
       Mix.Task.run("docs", args)
 
-      unless File.exists?("docs/index.html") do
-        Mix.raise "File not found: docs/index.html"
+      # make sure documentation directory exists
+      folder(true)
```
josevalim: It feels like adding the `print` parameter to this function is not giving us much, just making it a bit more complex. Wouldn't it be better to just check `if File.exists?("docs")` and the print the warnings directly here? :)

josevalim:
Added one comment, it looks great to me!

lib/mix/tasks/hex/docs.ex
JamesS237: That's a good point. I've followed your suggestions, squashed and pushed.

PR #77, opened 2015-01-03
Make warning about excluded dependencies more obvious

### LINKED ISSUE #70: Print warnings and dependencies when publishing package at the end
Today, a lot of people miss the warning where we say git packages will not be included. I propose for us to move the dependencies at the end (dependencies are more important than the package files in my opinion) and in case of warnings like this one or in case of missing fields, maybe have a [WARNING] or something in red prefixing it.

JamesS237:
Closes #70 

Preview:
screen shot 2015-01-03 at 3 44 38 pm

### DIFF
--- a/lib/mix/tasks/hex/publish.ex
+++ b/lib/mix/tasks/hex/publish.ex
@@ -122,16 +122,11 @@ defmodule Mix.Tasks.Hex.Publish do
       end)
     end
 
-    if exclude_deps != [] do
-      Mix.shell.info("  Excluded dependencies (not part of the Hex package):")
-      Enum.each(exclude_deps, &Mix.shell.info("    #{&1}"))
-    end
-
     if meta[:files] != [] do
       Mix.shell.info("  Included files:")
       Enum.each(meta[:files], &Mix.shell.info("    #{&1}"))
     else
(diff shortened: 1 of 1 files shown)


ericmj:
I like that we color it for visibility but can we make the color yellow and not bold since it's just a warning? Can we also color the metadata warning? :heart:

JamesS237:
@ericmj Done and done! :heart: 
screen shot 2015-01-03 at 4 45 13 pm

josevalim:
We unfortunately need to be careful with yellow because it doesn't work on white terminals: https://github.com/elixir-lang/elixir/commit/df44ad4fb2f8caf607e00d05147947d23ed8e1a3

If we remove bright it may just work™ but we would need to check to be double sure (or just go with red).

JamesS237:
It's working for me, but I removed the bright attribute when I switched to yellow ([see line 256](https://github.com/hexpm/hex/pull/77/files#diff-7617e617c077ac5a8780a3ea7a14ce95R256)), so I think we'll be good to go :tada: 
screen shot 2015-01-03 at 8 14 54 pm

josevalim:
Sweet!

PR #83, opened 2015-01-21
Add 'mix hex.user reset'; remove 'mix hex.user update'

jcspencer:
Related to hexpm/hex_web#89;

Changes:
- Adds the mix task `mix hex.user reset`. This accepts either a username *or* email address belonging to an account.
- Removes `mix hex.user reset`

:closed_lock_with_key: :key:

### DIFF
--- a/lib/mix/tasks/hex/user.ex
+++ b/lib/mix/tasks/hex/user.ex
@@ -31,7 +31,7 @@ defmodule Mix.Tasks.Hex.User do
 
   ### Reset user password
 
-  `mix hex.user update`
+  `mix hex.user reset password`
   """
 
   @switches [clean_pass: :boolean]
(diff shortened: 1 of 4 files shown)


ericmj:
We can remove `hex.user update` for now, but we need to add back the functionality to change email eventually. Changing email should basically go through the same flow as resetting password.

JamesS237:
@ericmj I agree. :+1:

jcspencer:
@ericmj Is this good to go now that password reset's are deployed? If not, what needs to be fixed?

ericmj:
@jcspencer I will take a closer in the next day. Thanks for all your great work!

jcspencer:
You're welcome! Thanks for the quick response :)

lib/mix/tasks/hex/user.ex
```diff
@@ -51,11 +51,11 @@ defmodule Mix.Tasks.Hex.User do
         create_key(opts)
       ["deauth"] ->
         deauth()
-      ["update"] ->
-        update(opts)
+      ["reset"] ->
```
ericmj: Should we change to "reset password" since we are probably adding "reset email" in the future or should we call it something else?

jcspencer: That sounds good to me, it'd be more suitable in the future.

jcspencer: Fixed :smile:

ericmj:
@jcspencer Looks great, thanks!

PR #111, opened 2015-07-17
Invalid ca bit

benmmurphy:


benmmurphy:
so this setting the user supplied cert as trusted_ca is fraught with danger but appears to be the best solution. my initial implementation only compared the public_key value in the cert which is very wrong because an attacker could swap in a different algorithm. it now checks subject public key info. hopefully, there is not a problem with this.

mix.exs
```diff
@@ -14,7 +14,7 @@ defmodule Hex.Mixfile do
   end
 
   defp deps do
-    [{:hex_web, github: "hexpm/hex_web", only: :test, env: :test}]
+    [{:hex_web, github: "hexpm/hex_web", only: :test, env: :test}, {:meck, github: "eproxus/meck", only: :test}]
```
lexmag: Is the any way to make testing without `:meck` usage?

benmmurphy: yeah. i could have a partial_chain/2 function that takes in a list of certs and make that public. then i just use this function directly with my list of certs and partial_chain/1 just calls partial_chain/2 with Hex.API.Certs.cacerts. or alternatively, if there is a way to modify the @der_encoded attribute on the Certs module then i can just modify this during the tests. i don't really have a good grasp of elixir so i don't know if this is possible.

lexmag: I think public `partial_chain/2` function is acceptable way (just `@doc false` it then).
`@der_encoded` can be modified but only before module will be compiled, usually with `__before_compile__/1`.

ericmj: I'll figure out a way to get rid of meck and will try this out locally. Thank you so much for this patch!

ericmj:
Ping @benmmurphy.

We are getting test failures because the supplied certificates are outdated. How can we fix this?

ericmj:
Ping? @benmmurphy

benmmurphy:
I'm sorry I just need to regen the certs. I'll have a look this weekend.

ericmj:
Thank you. When you do that can you drop a comment with the code on how to generate the certificates? Would it also be possible to generate without an expiration date?

PR #118, opened 2015-08-04
warn if files are missing

montague:
Addresses #99

### DIFF
--- a/lib/mix/tasks/hex/publish.ex
+++ b/lib/mix/tasks/hex/publish.ex
@@ -95,6 +95,7 @@ defmodule Mix.Tasks.Hex.Publish do
     package = Enum.into(config[:package] || [], %{})
 
     {deps, exclude_deps} = dependencies(config)
+    missing_files = missing_files(package[:files] || @default_files)
 
     meta = Keyword.take(config, [:app, :version, :elixir, :description])
            |> Enum.into(%{})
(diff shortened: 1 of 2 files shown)


mix.lock
```diff
@@ -1,15 +1,15 @@
 %{"aws_auth": {:hex, :aws_auth, "0.2.3"},
-  "bcrypt": {:git, "https://github.com/opscode/erlang-bcrypt.git", "7515e80a5e16f62aef149709f6cc09995dd480be", []},
+  "bcrypt": {:git, "git://github.com/opscode/erlang-bcrypt.git", "7515e80a5e16f62aef149709f6cc09995dd480be", []},
```
montague: i'm not sure why this change happened. i checked it in to make sure it came up in PR review, instead of just resetting the mix.lock file to reflect what was in master.

@ericmj any idea why running `mix deps.get` would do this?

whatyouhide:
@montague hey, I gave this a quick shot last week and the solution you adopted falls short when you have wildcards specified in the `:files` option (there are a bunch in the default files, like `readme*`) as `File.exists?/1` will return false for them. @ericmj suggested to warn for wildcards that match no files. What do you think?

montague:
sounds good. thanks for taking a look. i will add that.

mix.lock
ericmj: You are using an elixir version that creates git:// urls for github dependencies, later versions use https://. You can just skip commiting mix.lock.

montague:
We're now warning when a file or wildcard is missing.

I debated adding a separate message to distinguish between a missing file and a missing wildcard, but in the end I opted to keep it simple.

ericmj:
The missing files cannot be added to the metadata because that will be included in the package metadata.

montague:
i wondered about that. will refactor.

montague:
refactor complete.

ericmj:
Looks good, thank you!

PR #129, opened 2015-08-18
add build task

montague:
Add new `build` task to address #120.

Refactor `build` and `publish` tasks to use new `Mix.Hex.Build` module.

### DIFF
--- a/lib/mix/hex/build.ex
+++ b/lib/mix/hex/build.ex
@@ -1,0 +1,210 @@
+defmodule Mix.Hex.Build do
+  @default_files ~w(lib priv mix.exs README* readme* LICENSE*
+                    license* CHANGELOG* changelog* src)
+
+  @error_fields ~w(files app name version build_tools)a
+  @warn_fields ~w(description licenses contributors links)a
+  @meta_fields @error_fields ++ @warn_fields ++ ~w(elixir)a
+
+  def prepare_package! do
+    Hex.start
+    Hex.Utils.ensure_registry(update: false)
+    Mix.Project.get!
+
+    config = Mix.Project.config
+    raise_if_umbrella_project!(config)
+
+    package = Enum.into(config[:package] || [], %{})
+
+    {deps, exclude_deps} = dependencies(config)
+
+    meta = meta_for(config, package, deps)
+
+    [config: config, package: package, deps: deps,
+      exclude_deps: exclude_deps, meta: meta]
+  end
+
+  def meta_for(config, package, deps) do
+    Keyword.take(config, [:app, :version, :elixir, :description])
+    |> Enum.into(%{})
+    |> Map.merge(package)
+    |> package(config)
+    |> Map.put(:requirements, deps)
+  end
+
+  def print_info(meta, exclude_deps, package_files) do
+    if meta[:requirements] != [] do
+      Hex.Shell.info("  Dependencies:")
+      Enum.each(meta[:requirements], fn {app, %{requirement: req, optional: opt}} ->
+        message = "    #{app} #{req}"
+        if opt, do: message = message <> " (optional)"
+        Hex.Shell.info(message)
+      end)
+    end
+
+    Enum.each(@meta_fields, &print_meta(meta, &1))
+
+    warn_missing(meta)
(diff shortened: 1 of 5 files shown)


montague:
would love any feedback on this @ericmj @josevalim @lexmag

lib/mix/hex/utils.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Mix.Tasks.Hex.Utils do
+defmodule Mix.Hex.Utils do
```
lexmag: :+1: could it be in separate commit though?

lexmag:
It seems possible to avoid code duplication in `hex.publish` and `hex.build` tasks by running `build` task inside `publish` one.

montague:
good point. i will look into that tomorrow.

lib/mix/hex/utils.ex
montague: to confirm--you're saying the addition of that file should be in its own commit? sure.

lexmag: Yes, this improvement is clearly not related to the main work of PR. It will make easier to reason about in git history. I'd say it can even go into another PR, to reduce changes for review in the current.
As many git-style guides says:
> Don't make several logical changes in one commit.

montague: makes sense, thanks for clarifying.

ericmj:
Please rebase and force push instead of closing the PR.

montague:
will do--finishing up the changes soon.

lib/mix/hex/build.ex
```diff
@@ -0,0 +1,210 @@
+defmodule Mix.Hex.Build do
+  @default_files ~w(lib priv mix.exs README* readme* LICENSE*
+                    license* CHANGELOG* changelog* src)
+
+  @error_fields ~w(files app name version build_tools)a
+  @warn_fields ~w(description licenses contributors links)a
+  @meta_fields @error_fields ++ @warn_fields ++ ~w(elixir)a
+
+  def prepare_package! do
```
montague: I added this to further reduce duplication between `publish` and `build`. I think being able to simply call the `build` task inside the `publish` task (and then push to hex web) would require significantly more refactoring. For example, publishing and reverting both require access to the meta information. If we turn `--revert` into its own mix task, I think it'd be more straightforward to call the `build` task inside `publish`.

lib/mix/tasks/hex/publish.ex
```diff
@@ -76,41 +77,22 @@ defmodule Mix.Tasks.Hex.Publish do
...
-           |> Map.merge(package)
-           |> package(config)
-           |> Map.put(:requirements, deps)
-
-    if Mix.Project.umbrella?(config) do
-      Mix.raise "Hex does not support umbrella projects"
-    end
-
     if version = opts[:revert] do
       revert(meta, version, auth)
```
montague: I think pulling this into its own mix task would make it easier to refactor the `publish` task to simply call `build` and then push to hex web.

montague:
I reverted the namespace change of `Mix.Tasks.Hex.Utils` to avoid cluttering up this PR. I'll reimplement that in a follow-up PR when this one gets merged.

montague:
As always, thanks everyone for taking a look!

lib/mix/tasks/hex/build.ex
```diff
@@ -0,0 +1,77 @@
+defmodule Mix.Tasks.Hex.Build do
+  use Mix.Task
+  alias Mix.Tasks.Hex.Utils
+  alias Mix.Hex.Build
+
+  @shortdoc "Build a new package version"
+
+  @moduledoc """
+  Build a new version of your package.
```
ericmj: Can we clarify what this does in another paragraph? `mix compile` also builds the package so it's not clear what this actually does.

Mention that it works like `hex.publish` but it only builds the package tarball into the local directory, what name it will have, and that it doesn't actually publish to a repository.

ericmj:
Looks good. I dropped a small comment where I think documentation can be improved.

montague:
Docs updated

ericmj:
Thank you!

PR #173, opened 2015-12-22
fixes #172

### LINKED ISSUE #172: Hex should not choke on invalid requirements
The following `mix.exs` crashes Hex:

```elixir
defmodule ISO.Mixfile do
  use Mix.Project

  def project do
    [
      app: :iso,
      version: "0.1.0",
      deps: deps
    ]
  end

  def application do
    [
      applications: [:logger]
    ]
  end

  defp deps do
    [
      { :phoenix,  "" },
    ]
  end
end
```

RobertDober:
I guess the label starter was justified ;).
Oh and about the message, in `Mix.raise` I guess it is lousy, please improve on it anyone :blush:

### DIFF
--- a/lib/hex/remote_converger.ex
+++ b/lib/hex/remote_converger.ex
@@ -105,13 +105,17 @@ defmodule Hex.RemoteConverger do
   end
 
   defp check_package_req(name, req, from) do
-    if versions = Registry.get_versions(name) do
-      versions = Enum.filter(versions, &Hex.Mix.version_match?(&1, req))
-      if versions == [] do
-        Mix.raise "No package version in registry matches #{name} #{req} (from: #{from})"
+    try do
+      if versions = Registry.get_versions(name) do
+        versions = Enum.filter(versions, &Hex.Mix.version_match?(&1, req))
+        if versions == [] do
+          Mix.raise "No package version in registry matches #{name} #{req} (from: #{from})"
+        end
+      else
+        Mix.raise "No package with name #{name} (from: #{from}) in registry"
       end
-    else
-      Mix.raise "No package with name #{name} (from: #{from}) in registry"
+    catch _, %Version.InvalidRequirementError{message: message} ->
+      Mix.raise "#{name}'s version is not correctly specified, version: `#{message}`"
     end
   end
 
--- a/test/hex/mix_task_test.exs
+++ b/test/hex/mix_task_test.exs
@@ -103,6 +103,14 @@ defmodule Hex.MixTaskTest do
     end
   end
 
+  defmodule WithEmptyDepVersion do
+    def project do
+      [ app: :simple,
+        version: "0.1.0",
+        deps: [ {:ecto, ""} ] ]
+    end
+  end
+
   test "deps.get" do
     Mix.Project.push Simple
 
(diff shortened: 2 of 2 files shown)


lib/hex/remote_converger.ex
```diff
@@ -105,13 +105,17 @@ defmodule Hex.RemoteConverger do
   end
 
   defp check_package_req(name, req, from) do
-    if versions = Registry.get_versions(name) do
-      versions = Enum.filter(versions, &Hex.Mix.version_match?(&1, req))
-      if versions == [] do
-        Mix.raise "No package version in registry matches #{name} #{req} (from: #{from})"
+    try do
```
lexmag: I think it needs to be solved without using `try`.
It depends on how we see `""` requirement.
If it should fail – we could add extra `check_package_req` clause, if it is the same as `nil` – extra `Hex.Mix.version_match?` clause then.

test/hex/mix_task_test.exs
```diff
@@ -418,4 +426,17 @@ defmodule Hex.MixTaskTest do
...
+    Mix.Project.push WithEmptyDepVersion
+
+    in_tmp fn ->
+      Hex.State.put(:home, System.cwd!)
+
+      assert_raise Mix.Error, "ecto's version is not correctly specified, version: ``", fn ->
+        Mix.Task.run "deps.get"
+      end
+    end
+
```
lexmag: Needless line :scissors:.

lexmag:
@RobertDober please adjust commit message such that it describes what it changes without relying on GitHub issue link. :bowtie:

lib/hex/remote_converger.ex
lexmag: However, considering other possible invalid definitions of requirement we may need this `try`.

lexmag: We need make `Hex.Mix.version_match?` to return `:ok` | `:error` and move `try` in it.

RobertDober:
@lexmag I too wanted to avoid try. However it seemed complicated at first sight. Maybe I will give it another try (pun intended) by checking with `Version.parse` inside `Registry.get_versions` returning {:error, message} and then replacing the if with a case inside `check_package_req`.

What would be a good commit message for PRs. I guess I shall have a look at some other PRs.

lib/hex/remote_converger.ex
lexmag: Actually the easiest way would be to use `Version.parse_requirement` before `versions = Enum.filter...`.

RobertDober: That seems the best. Some extra work to catch all incorrect values maybe...
Ty for all these ideas.

RobertDober: Now I am confused. :confused:  It is not `versions` that contain the illegal version spec, but `req`, correct?

Or did I get the test setup wrong?

RobertDober: Ok you gave me the correct hint about Hex.Mix.version_match? we will just do it with Version.parse_requirement instead of try. I'll come back with a new PR before this becomes too confusing.

lexmag: @RobertDober yes the error comes from invalid `req`.

RobertDober: found an easy solution, PR coming up

PR #174, opened 2015-12-22
[Solution#172] raises a Mix.Error instead of choking on invalid requi…

RobertDober:
…rements (only fot hex requirements)

Seems easy enough after all, thx to @lexmag for clearing things up.

### DIFF
--- a/lib/hex/remote_converger.ex
+++ b/lib/hex/remote_converger.ex
@@ -105,6 +105,9 @@ defmodule Hex.RemoteConverger do
   end
 
   defp check_package_req(name, req, from) do
+    if req && Version.parse_requirement(req) == :error do
+        Mix.raise "Required version #{inspect req} for package #{name} is incorrectly specified (from: #{from})"
+    end
     if versions = Registry.get_versions(name) do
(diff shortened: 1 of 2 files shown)


lib/hex/remote_converger.ex
```diff
@@ -105,6 +105,9 @@ defmodule Hex.RemoteConverger do
   end
 
   defp check_package_req(name, req, from) do
+    if req && Version.parse_requirement(req) == :error do
```
lexmag: `nil` requirement is still valid.

lexmag:
> What would be a good commit message for PRs. I guess I shall have a look at some other PRs.

Regarding commit message, I think this guide is pretty good: https://github.com/erlang/otp/wiki/Writing-good-commit-messages.

lib/hex/remote_converger.ex
RobertDober: That is why I do not raise a Mix.Error in that case.

lexmag:
Thank you. :yellow_heart:

RobertDober:
And thanx to you too, Happy Xmas. :christmas_tree:

PR #178, opened 2016-01-15
change Hex.shell to Hex.Shell

sa1:
This seems to be a typo.


sa1:
oh, an identical PR exists. Closing.

PR #185, opened 2016-01-22
tests(Hex.Mix): add tests for overriding dependencies

Gazler:
The names of the dependencies have been renamed to actual libraries so
that the tests are easier to understand.


Gazler:
@ericmj You merged the last PR literally as I was writing this commit message.

I could use a review on the test to make sure it is correct anyway. :)

josevalim:
:shipit:

josevalim:
:shipit:

PR #188, opened 2016-01-24
Catch an empty search result

merimna:
Instead of throwing an Enum.EmptyError, return a message about no packages found.


lexmag:
Thank you!

PR #230, opened 2016-05-14
Warn about long package descriptions

### LINKED ISSUE #229: Warn if description is too long
See https://github.com/hexpm/hex_web/issues/326

rrrene:
Warns the user during `mix hex.publish` if the package description exceeds 300 characters.

Fixes #229 :grin:

### DIFF
--- a/lib/mix/hex/build.ex
+++ b/lib/mix/hex/build.ex
@@ -5,6 +5,7 @@ defmodule Mix.Hex.Build do
   @error_fields ~w(files app name description version build_tools)a
   @warn_fields ~w(licenses maintainers links)a
   @meta_fields @error_fields ++ @warn_fields ++ ~w(elixir)a
+  @max_description_length 300
 
   def prepare_package! do
     Hex.start
@@ -36,6 +37,7 @@ defmodule Mix.Hex.Build do
     Enum.each(@meta_fields, &print_meta(meta, &1))
 
     warn_missing(meta)
+    warn_long_description(meta)
     warn_missing_files(package_files)
     error_missing!(meta)
 
@@ -169,6 +171,14 @@ defmodule Mix.Hex.Build do
       fields = Enum.join(missing_fields, ", ")
       Hex.Shell.error("  ERROR! Missing metadata fields: #{fields}")
       Mix.raise("Stopping package build due to errors")
+    end
+  end
+
+  defp warn_long_description(meta) do
+    description = meta[:description] || ""
+
+    if String.length(description) > @max_description_length do
+      Hex.Shell.warn("  WARNING! Package description is very long (exceeds #{@max_description_length} characters)")
     end
   end
 
--- a/test/mix/tasks/hex/build_test.exs
+++ b/test/mix/tasks/hex/build_test.exs
@@ -88,4 +88,19 @@ defmodule Mix.Tasks.Hex.BuildTest do
   after
     purge [ReleaseNoDescription.Mixfile]
   end
+
+  test "warn if description is too long" do
+    Mix.Project.push ReleaseTooLongDescription.Mixfile
+
+    in_tmp fn ->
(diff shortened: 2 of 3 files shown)


lib/mix/hex/build.ex
```diff
@@ -172,6 +174,14 @@ defmodule Mix.Hex.Build do
     end
   end
 
+  defp warn_long_description(meta) do
+    description = meta |> Map.get(:description) |> to_string
```
lexmag: It'd be better to use `meta[:description] || ""` since we expect description to be either `nil` or string.

lib/mix/hex/build.ex
```diff
@@ -172,6 +174,14 @@ defmodule Mix.Hex.Build do
...
+
+    if String.length(description) > @max_description_length do
+      Hex.Shell.warn("  WARNING! Package description is very long (> #{@max_description_length}).")
```
lexmag: What if we say `"  WARNING! Package description is too long (exceeds #{@max_description_length} characters)."`?

rrrene:
@lexmag Good points, I updated the PR accordingly!

lib/mix/hex/build.ex
```diff
@@ -5,6 +5,7 @@ defmodule Mix.Hex.Build do
   @error_fields ~w(files app name description version build_tools)a
   @warn_fields ~w(licenses maintainers links)a
   @meta_fields @error_fields ++ @warn_fields ++ ~w(elixir)a
+  @max_description_length 300 # characters
```
whatyouhide: To be very nitpicky, these are *graphemes*, not characters, as pointed out by the [documentation for `String.length/1`](http://elixir-lang.org/docs/stable/elixir/String.html#length/1) (that we call below).

lib/mix/hex/build.ex
```diff
@@ -172,6 +174,14 @@ defmodule Mix.Hex.Build do
     end
   end
 
+  defp warn_long_description(meta) do
+    description = meta[:description] || ""
+
+    if String.length(description) > @max_description_length do
+      Hex.Shell.warn("  WARNING! Package description is very long (exceeds #{@max_description_length} characters).")
```
lexmag: Let's use `too long` instead of `very long`. :bowtie:

rrrene: Sure I can change it.

The reason I picked "very" over "too" is that the second seems to imply consequences. Since this just gives a warning, maybe another wording would be more precise. Something like "WARNING!   Package description is verbose (exceeds 300 characters)."

Thoughts?

lib/mix/hex/build.ex
rrrene: Maybe I should just ditch the comment then?

lib/mix/hex/build.ex
lexmag: Good point about "this just gives a warning", `too` feels better for an error and `very` for a warning. 👍

lib/mix/hex/build.ex
lexmag: Probably yes, I think `_length` shows what we use for a measuring (`String.length/1`).

rrrene:
Updated again.

I re-read [the discussion](https://github.com/hexpm/hex_web/issues/326) which created #229 - if we actually truncate the description at 300 graphemes before saving it to the registry (in the future), then we might want to reconsider the wording of the warning and whether or not this should throw an error and prevent publication.

But IMHO the warning does improve the situation (without requiring server-side modifications).

lib/mix/hex/build.ex
lexmag: Very last nitpick and we're good to merge: we don't use trailing period in warning messages.

rrrene:
@lexmag trailing point is gone.

lexmag:
@rrrene thank you!

PR #294, opened 2016-09-24
Polish the deprecation message in "mix hex.docs"

whatyouhide:


whatyouhide:
@ericmj whoops, completely missed that test to update. Updated now :)

PR #301, opened 2016-10-22
Fix location of generated html files

wsmoak:
By default the `mix docs` task generates the html output into `doc/` (no s).


wojtekmach:
Nice catch, thanks!

PR #336, opened 2016-12-26
Reduce nested code in some functions

milmazz:


lib/hex/resolver/backtracks.ex
```diff
@@ -93,19 +93,19 @@ defmodule Hex.Resolver.Backtracks do
...
-          case try_merge_package_version(list, version, parents) do
-            {:ok, list} ->
-              Map.put(map, name, list)
-            :error ->
-              Map.put(map, name, [{[version], parents}|list])
-          end
-
+         list
+         |> try_merge_package_version(version, parents)
+         |> process_merge_package_version(map, name, version, parents, list)
```
lexmag: I think previously code was clearer, new function takes too many arguments.

lib/hex/remote_converger.ex
```diff
@@ -149,17 +149,17 @@ defmodule Hex.RemoteConverger do
...
-          retired ->
-            Hex.Shell.warn "  #{name} #{version} RETIRED!"
-            Hex.Shell.warn "    (#{retirement_reason(retired[:reason])}) #{retired[:message]}"
-        end
+        name |> Registry.retired(version) |> show_status(name, version)
       end)
     end
   end
 
+  defp show_status(nil, name, version), do: Hex.Shell.info IO.ANSI.format [:green, "  #{name} #{version}"]
```
lexmag: We usually prefix such functions with `print_` instead of `show_`.

lexmag: Also there is too much things are happening inside this clause, I think we should use `do ... end` syntax.

lib/hex/remote_converger.ex
```diff
@@ -173,26 +173,31 @@ defmodule Hex.RemoteConverger do
...
+          verify_registry(deps, name, version, checksum)
         _ ->
           :ok
       end
     end)
   end
 
+  defp verify_registry(deps, name, version, checksum) do
+    registry_checksum = Registry.checksum(name, version)
+    if checksum && Base.decode16!(checksum, case: :lower) != registry_checksum,
```
lexmag: `do ... end` syntax would be preferable here.

lib/hex/remote_converger.ex
```diff
@@ -173,26 +173,31 @@ defmodule Hex.RemoteConverger do
...
+    if deps, do: verify_dependencies(deps, name, version)
+  end
+
+  defp verify_dependencies(deps, name, version) do
+    deps =
+      Enum.map(deps, fn {app, req, opts} ->
+        {Atom.to_string(opts[:hex]), Atom.to_string(app), req, !!opts[:optional]}
+      end)
+
+    if Enum.sort(deps) != Enum.sort(Registry.deps(name, version)),
```
lexmag: `do ... end` syntax would be preferable here.

lib/hex/crypto/pbes2_hmac_sha2.ex
```diff
@@ -58,6 +53,9 @@ defmodule Hex.Crypto.PBES2_HMAC_SHA2 do
   end
   def decrypt(_, _, _, _), do: :error
 
+  defp process_fetch_p2s({:ok, salt}, hash, passwd), do: {:ok, %{hash: hash, password: passwd}}
```
lexmag: Maybe `handle_p2s`? `process_fetch_p2s` seems misleading.

lib/hex/crypto/key_manager.ex
```diff
@@ -28,12 +28,9 @@ defmodule Hex.Crypto.KeyManager do
         case module.init(protected, opts) do
           {:ok, params} ->
             key_manager = %KeyManager{module: module, params: params}
-            case ContentEncryptor.init(protected, opts) do
-              {:ok, content_encryptor} ->
-                {:ok, key_manager, content_encryptor}
-              content_encryptor_error ->
-                content_encryptor_error
-            end
+            protected
```
lexmag: We can probably have `fetch_content_encryptor` function instead:
```elixir
defp fetch_content_encryptor(key_manager, protected, opts) do
  case ContentEncryptor.init(protected, opts) do
    {:ok, content_encryptor} ->
      {:ok, key_manager, content_encryptor}
    error ->
      error
end
```

lib/hex/crypto/pbes2_hmac_sha2.ex
milmazz: Thanks, I was looking for a better name in this function...

lib/hex/crypto/key_manager.ex
milmazz: That also works, I'll change this section as you suggested.

milmazz:
> my only minor comment is that we should have only one pipe per line

@ericmj Already take care of that in the latest commit.

@lexmag Thanks for your review, please let me know what do you think about the latest commit.

Issue #363, opened 2017-04-19
mix hex.repo show NAME


Angry-Potato:
The `mix hex.repo show NAME` task is not implemented.

Returns 
```
** (Mix) Invalid arguments, expected one of:
mix hex.repo add NAME URL [PUBLIC_KEY_PATH [AUTH_KEY]]
mix hex.repo set url NAME URL
mix hex.repo set public_key NAME PUBLIC_KEY_PATH
mix hex.repo set auth_key NAME AUTH_KEY
mix hex.repo remove NAME
mix hex.repo show NAME
mix hex.repo list
```
regardless of arguments

PR #373, opened 2017-05-10
Add hex.repo subtask to show repo config

parihar:
Fixes https://github.com/hexpm/hex/issues/363.

When repo is found:
```sh
$ mix hex.repo show hexpm
URL                  Public key                                          Auth key
https://repo.hex.pm  SHA256:O1LOYhHFW4kcrblKAxROaDEzLD8bn1seWbe5tq8TRsk
```

When repo is not found:
```sh
$ mix hex.repo show somerepo
** (Mix) Config does not contain repo somerepo
```

### DIFF
--- a/test/mix/tasks/hex.repo_test.exs
+++ b/test/mix/tasks/hex.repo_test.exs
@@ -1,8 +1,6 @@
 defmodule Mix.Tasks.Hex.RepoTest do
   use HexTest.Case
 
-  import ExUnit.CaptureIO
-
   @public_key """
   -----BEGIN PUBLIC KEY-----
   MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEApqREcFDt5vV21JVe2QNB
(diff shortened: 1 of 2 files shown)


test/mix/tasks/hex.repo_test.exs
```diff
@@ -48,4 +48,23 @@ defmodule Mix.Tasks.Hex.RepoTest do
       assert ["$repos": %{"reponame" => %{url: "other_url"}}] = Hex.Config.read
     end
   end
+
+  test "show returns repo config" do
+    in_tmp fn ->
+      Hex.State.put(:home, System.cwd!)
+
+      Mix.Tasks.Hex.Repo.run(["add", "reponame", "url"])
+      assert %{url: "url"} = Mix.Tasks.Hex.Repo.run(["show", "reponame"])
```
ericmj: We shouldn't return things from tasks. Technically tasks are specced to only return `:ok` so we need to test the output with `assert_received`.

parihar:
I wasn't sure how to test with `assert_received`, so tried with `capture_io`.

wojtekmach:
Please see eg
https://github.com/parihar/hex/blob/664b7a5947f495cdf8dd5c98032dbe068c566bab/test/mix/tasks/hex.info_test.exs
how to use assert_receive

W dniu śr., 10.05.2017 o 12:53 Abhi Parihar <notifications@github.com>
napisał(a):

parihar:
`assert received` is definitely cleaner. Would you like these commits squashed?

parihar:
Cool. Thanks for the help.

PR #390, opened 2017-06-24
Fix persisting registry on Elixir < 1.4.0

wojtekmach:
Without this, there's a deadlock on Elixir 1.3 where both `Hex.SCM` and `Hex.RemoteConverger` are trying to persist.

### DIFF
--- a/lib/hex/registry/server.ex
+++ b/lib/hex/registry/server.ex
@@ -111,14 +111,9 @@ defmodule Hex.Registry.Server do
     {:noreply, state}
   end
 
-  def handle_call(:persist, from, state) do
-    state = wait_closing(state, fn ->
-      persist(state.ets, state.path)
-      GenServer.reply(from, :ok)
-      state
-    end)
-
-    {:noreply, state}
+  def handle_call(:persist, _from, state) do
+    persist(state.ets, state.path)
+    {:reply, :ok, state}
   end
 
   def handle_call({:prefetch, packages}, _from, state) do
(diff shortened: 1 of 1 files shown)


lib/hex/registry/server.ex
```diff
@@ -111,14 +111,9 @@ defmodule Hex.Registry.Server do
...
-    state = wait_closing(state, fn ->
-      persist(state.tid, state.path)
-      GenServer.reply(from, :ok)
-      state
-    end)
-
-    {:noreply, state}
+  def handle_call(:persist, _from, state) do
+    persist(state.ets, state.path)
+    {:reply, :ok, state}
```
ericmj: Hum, this is tricky, because we should wait until all responses have arrived until persisting otherwise we can miss persisting some responses. This means we currently have to wait until all responses have arrived until we can start checking out packages in Hex.SCM but I guess we have to live this on <1.4.

I see that travis is failing because I wrote `state.ets` instead of `state.tid` but do we need to remove the wait_closing also? I think we need to find another solution.

wojtekmach:
I have 2 commits on this PR, the 1st fixes the typo but then the tests
hang, thus the 2nd commit - try running locally on 1.3.4 without 2nd one
W dniu niedz., 25.06.2017 o 13:12 Eric Meadows-Jönsson <
notifications@github.com> napisał(a):

ericmj:
> I have 2 commits on this PR, the 1st fixes the typo but then the tests
hang, thus the 2nd commit - try running locally on 1.3.4 without 2nd one

Okay, I will try that, but I do think we need to find another fix.

wojtekmach:
Agreed it was a hack to get the build passing
W dniu niedz., 25.06.2017 o 13:21 Eric Meadows-Jönsson <
notifications@github.com> napisał(a):

ericmj:
I went with your approach in https://github.com/hexpm/hex/pull/396. I think it's the best we can do until we figure out something better. Sorry for not merging earlier.

PR #400, opened 2017-07-31
[hex.search] use package html_url

asgoel:


wojtekmach:
Thanks @asgoel. ON master, I've added the same `package["html_url"] || package["url"]` fallback as in the other place.

PR #401, opened 2017-07-31
Make html_url optional

wojtekmach:
In the spec [1] we've made `html_url` required, however I can see it being optional too. I imagine someone could build a hex custom repo with just APIs and no HTML stuff. If we make it optional, we still need to display some URL when package is published, so I think falling back to API url is reasonable.

This really is only an issue when we allow publishing packages to non-hexpm repos, however leaving it here as a mental note :)

[1] https://github.com/hexpm/specifications/pull/21/files#diff-2625016b50d68d922257f74801cac29cR313

cc @asgoel


asgoel:
good call. The minimal hex repo I've been working on doesn't have HTML support yet - I was just planning on making `html_url` = `api_url` but this works better :)

PR #409, opened 2017-08-23
Use MixProject instead of Mixfile in mix.exs files

gfvcastro:
Mix renames the module in `mix.exs` file from `Mixfile` to `MixProject`.

This pull request updates the name.

Reference: https://github.com/elixir-lang/elixir/commit/c6a0edcd32359eca284ff0166d19a6dc979ee954


wojtekmach:
Thank you!

PR #415, opened 2017-09-05
Add description to hex.search task

gfvcastro:


lib/hex/utils.ex
```diff
@@ -81,6 +81,18 @@ defmodule Hex.Utils do
...
 
+  def truncate(string, options \\ []) do
+    length = options[:length] || 50
+    omission = options[:omission] || "..."
+
+    cond do
+      !String.valid?(string) -> string
+      String.length(string) < length -> string
+      true ->
+        String.slice(string, 0, length) <> omission
```
ericmj: Please be consistent with new lines for clauses.

lib/hex/utils.ex
```diff
@@ -81,6 +81,18 @@ defmodule Hex.Utils do
     safe_tuple(tuple, n - 1)
   end
 
+  def truncate(string, options \\ []) do
+    length = options[:length] || 50
+    omission = options[:omission] || "..."
+
+    cond do
+      !String.valid?(string) -> string
```
ericmj: Use `not` instead of `!` when you know the value is a boolean.

wojtekmach:
Cool idea!

Turns out some packages (e.g. [1]) have `\n` in their description which breaks the output:

zrzut ekranu 2017-09-06 o 19 27 52

Could you handle that edge case? I think breaking on first `\n` and then making sure it's less than 50 should do the trick.

[1] https://github.com/romul/airbrake-elixir/blob/master/mix.exs

gfvcastro:
Good catch! Thank you!

ericmj:
Thank you!

ericmj:
Thank you!

PR #421, opened 2017-09-11
Escape owner email address when adding or deleting an owner.

chingor13:
An email address can contain invalid path uri components such as "+". If you
try to add or remove an email address with a "+" in it, you will get
back a 404.


wojtekmach:
Thank you!

PR #422, opened 2017-09-12
Add "integration" tag to hex.organization_test.exs

parkerl:
It appears this test file may be missing the "integration" tag. It fails when run without the local server up.


wojtekmach:
Thank you!

PR #441, opened 2017-10-19
Preserve symlinks and empty directories in tar

mobileoverlord:
Empty directories and symlinks are currently omitted when packaging the tar file. This PR allows empty directories and symlinks to be retained. 

This should be rebased once https://github.com/hexpm/hex/pull/440 lands

### DIFF
--- a/lib/hex/tar.ex
+++ b/lib/hex/tar.ex
@@ -38,9 +38,14 @@ defmodule Hex.Tar do
         {name, contents} ->
           :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), [])
         name ->
-          contents = File.read!(name)
-          mode = File.stat!(name).mode
-          :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), mode, [])
+          case File.lstat(name) do
+            {:ok, %File.Stat{type: type}} when type in [:directory, :symlink] ->
+              :ok == :hex_erl_tar.add(tar, Hex.string_to_charlist(name), [])
+            _stat ->
+              contents = File.read!(name)
+              mode = File.stat!(name).mode
+              :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), mode, [])
+          end
       end)
     after
       :hex_erl_tar.close(tar)
--- a/lib/hex/tar.ex
+++ b/lib/hex/tar.ex
@@ -38,7 +38,7 @@ defmodule Hex.Tar do
         {name, contents} ->
           :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), [])
         name ->
-          case File.lstat(name) do
+          case Hex.file_lstat(name) do
             {:ok, %File.Stat{type: type}} when type in [:directory, :symlink] ->
               :ok = :hex_erl_tar.add(tar, Hex.string_to_charlist(name), [])
             _stat ->
--- a/lib/hex/tar.ex
+++ b/lib/hex/tar.ex
@@ -40,7 +40,7 @@ defmodule Hex.Tar do
         name ->
           case File.lstat(name) do
             {:ok, %File.Stat{type: type}} when type in [:directory, :symlink] ->
-              :ok == :hex_erl_tar.add(tar, Hex.string_to_charlist(name), [])
+              :ok = :hex_erl_tar.add(tar, Hex.string_to_charlist(name), [])
             _stat ->
               contents = File.read!(name)
               mode = File.stat!(name).mode
--- a/lib/mix/tasks/hex.build.ex
+++ b/lib/mix/tasks/hex.build.ex
@@ -367,7 +367,7 @@ defmodule Mix.Tasks.Hex.Build do
   end
 
   defp dir_files(path) do
-    case File.lstat(path) do
+    case Hex.file_lstat(path) do
       {:ok, %File.Stat{type: :directory}} ->
         [path | Path.wildcard(Path.join(path, "**"), match_dot: true)]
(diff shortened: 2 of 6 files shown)


lib/mix/tasks/hex.build.ex
```diff
@@ -267,16 +267,16 @@ defmodule Mix.Tasks.Hex.Build do
     |> Enum.flat_map(&Path.wildcard/1)
     |> Enum.flat_map(&dir_files/1)
     |> Enum.map(&Path.expand/1)
-    |> Enum.filter(&File.regular?/1)
```
mobileoverlord: Removing this line makes me question what types of non-regular files we are interested in omitting.

lib/hex/tar.ex
```diff
@@ -38,9 +38,14 @@ defmodule Hex.Tar do
         {name, contents} ->
           :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), [])
         name ->
-          contents = File.read!(name)
-          mode = File.stat!(name).mode
-          :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), mode, [])
+          case File.lstat(name) do
+            {:ok, %File.Stat{type: type}} when type in [:directory, :symlink] ->
```
ericmj: Won't this break for symlinks since we are not storing the target of the symlink? Maybe we should remove those, I don't know if `:erl_tar` supports symlinks?

lib/mix/tasks/hex.build.ex
ericmj: I guess I just didn't see any reason for storing empty directories at the time.

lib/hex/tar.ex
ericmj: Never mind, I just saw how you call `:erl_tar` below.

ericmj: I believe that call below will add the current mtimes to the tar headers though which will break reproducibility. The reason we use `:hex_erl_tar.add/5` instead of `:hex_erl_tar.add/4` is so we don't preserve mtimes.

We don't seem to have tests that verifies this so can you please add assertions to the test that verifies we don't preserve mtimes for files or directories. If the test fails I will look into how we can solve this.

mobileoverlord:
I rebased this and added checks for mtime to the test. It started to fail now. If I understand correctly, the reason hex holds its own copy of erl_tar is to make the mtime modifications?

ericmj:
We vendor `erl_tar` to bring features from OTP 20 to all Hex users. Later we had to do modifications to support setting file modes while still setting mtime to unix epoch. I think we will have to do further modifications to support this use case, I will look into it today.

ericmj:
@mobileoverlord I pushed a few commits fixing the mtime issues for directories and symlinks and rebased your commits.

This is ready for review @wojtekmach.

lib/hex/tar.ex
```diff
@@ -38,9 +38,14 @@ defmodule Hex.Tar do
         {name, contents} ->
           :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), [])
         name ->
-          contents = File.read!(name)
-          mode = File.stat!(name).mode
-          :ok = :hex_erl_tar.add(tar, contents, Hex.string_to_charlist(name), mode, [])
+          case Hex.file_lstat(name) do
+            {:ok, %File.Stat{type: type}} when type in [:directory, :symlink] ->
+              :ok == :hex_erl_tar.add(tar, Hex.string_to_charlist(name), [])
```
wojtekmach: `:ok = :hex_erl_tar.add(...)`

src/hex_erl_tar.erl
```diff
@@ -1261,12 +1261,10 @@ convert_header(_Bin, _Reader) ->
 %% If the file is a directory, a slash is appended to the name.
 fileinfo_to_header(Name, #file_info{}=Fi, Link) when is_list(Name) ->
     BaseHeader = #tar_header{name=Name,
-                             mtime=Fi#file_info.mtime,
-                             atime=Fi#file_info.atime,
-                             ctime=Fi#file_info.ctime,
+                             mtime=0,
+                             atime=0,
+                             ctime=0,
```
wojtekmach: :+1:

btw, I noticed that it works well for files but not for dirs: (output from unpacked dummy package)

zrzut ekranu 2017-10-22 o 19 49 31

I'm gonna create a separate issue for it if it's worth pursuing.

wojtekmach:
@milmazz just wanted to thank you again for your work on `mix hex.build --smoke` - it made testing this particular PR so much easier ❤️

src/hex_erl_tar.erl
wojtekmach: the output above is when I manually did: `tar xvzf contents.tar.gz`. I did manual unpacking because I thought there's a bug in `mix hex.build --smoke` (which I tried first) because when it unpacked it looked like it didn't preserve timestamps. However, while we're trying to preserve timestamps _in the archive_ when we unpack we're touching files [1] so I guess that was by design. Still I wonder if we should be zeroing timestamps for dirs in the archive.

[1] https://github.com/hexpm/hex/blob/master/lib/hex/tar.ex#L115

ericmj: Yes, I have a comment about that in the tests. It would require bigger changes to erl_tar.

ericmj: I'm not sure if we need to fix it for extracting unless someone actually asks for it, if `tar` has the same behavior then maybe it's even working as intended. What's important to ensure is that the tarballs we build do not include bad mtimes.

milmazz:
@wojtekmach I'm glad you find it useful!

Issue #424, opened 2017-09-13
Improve display and warnings when publishing private packages


brweber2:
In order to publish a private package you run:
```
mix hex.publish --organization <org-name>
```

If you forget to add the --organization flag it publishes your private code **publicly**. I think that it should publish to the private repo by default if :organization is in mix.exs. If someone wishes to publish a private package publicly then they could use a flag to indicate they they do not want the default behavior. As a (future) paying customer I do not want my private packages ending up on hex.pm and if the mistake happens I only have one hour to undo it. So if I don't notice immediately my code will remain public until an admin removes it.

My short term fix for this has been to add an alias, but it seems like the default behavior is not "safe".

josevalim:
I agree that if you put an `:organization` under the `:package` key in the mix.exs then you should always publish under that organization.

ericmj:
@brweber2 Can you show your mix.exs file? The `:organization` key should be added under the `:package` key. It will not work if you add it directly under `def project`.

brweber2:
Ah, we have it directly in the keyword list in :project.  So that is where we went wrong, but I’m sure we won’t be the last group to do this. The docs are technically correct, but it is very subtle that the config shown is package and not project.

Thanks for clarifying this!

josevalim:
@ericmj given this can be a very sensitive issue, maybe it is worth checking for an `:organization` key being misplaced?

wojtekmach:
I think in this particular case it makes sense to raise an error if we detect misplaced organisation/organization (easy to make typo from different en speaking countries)

What we can also do is make it prominent to which repo we are publushing before hitting "Y", perhaps by using bold and/or colored text and putting the message in a separate line.

josevalim:
Yes, maybe say "PUBLICLY" on the regular publishing messages or even have a separate workflow for publishing for orgs... although that wouldn't necessarily address the muscle memory issue.

ericmj:
I agree with all points.

  * Check for British spelling `:organisation` and warn
  * Check if any options under `:package` are given under the root and warn
  * Make it more clear than it is today if you're pushing to the public repository or a private one

brweber2:
Ah, we have it directly in the keyword list in :project.  So that is where we went wrong, but I’m sure we won’t be the last group to do this. The docs are technically correct, but it is very subtle that the config shown is package and not project.

Thanks for clarifying this!

ericmj:
[...]

PR #450, opened 2017-10-29
Make repository more prominent in hex.publish

wojtekmach:
Ref #424

Before:

screen shot 2017-10-29 at 21 25 30

After:

screen shot 2017-10-29 at 21 24 04
screen shot 2017-10-29 at 21 24 26

### DIFF
--- a/lib/mix/tasks/hex.publish.ex
+++ b/lib/mix/tasks/hex.publish.ex
@@ -205,11 +205,19 @@ defmodule Mix.Tasks.Hex.Publish do
       not confirm? ->
         true
       build.organization in [nil, "hexpm"] ->
-        Hex.Shell.info("Publishing package to **public** repository hexpm.")
+        Hex.Shell.info(["Publishing package to ", emphasis("public"), " repository hexpm."])
         Hex.Shell.yes?("Proceed?")
       true ->
-        Hex.Shell.info("Publishing package to **private** repository #{build.organization}.")
+        Hex.Shell.info(["Publishing package to ", emphasis("private"), " repository #{build.organization}."])
         Hex.Shell.yes?("Proceed?")
+    end
+  end
+
+  defp emphasis(text) do
+    if IO.ANSI.enabled? do
+      [IO.ANSI.bright(), text, IO.ANSI.reset()]
+    else
+      ["**", text, "**"]
     end
   end
 
--- a/lib/mix/tasks/hex.publish.ex
+++ b/lib/mix/tasks/hex.publish.ex
@@ -205,14 +205,16 @@ defmodule Mix.Tasks.Hex.Publish do
       not confirm? ->
         true
       build.organization in [nil, "hexpm"] ->
-        Hex.Shell.yes?("Publishing package to public repository hexpm. Proceed?")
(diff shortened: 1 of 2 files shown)


josevalim:
It may be better to use bold and ** around the text. ANSI colors do not
work one windows and may be less readable depending on the background
color, so we need to be careful when relying on them for emphasis.
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

wojtekmach:
If there's a way to detect whether a shell supports escape codes then perhaps we could still use colours but fallback to text emphasis? I do agree that using these particular colours might be problematic for users with light background, will do some research.

What I wanted to achieve with using _different_ colours is to train users to treat cyan as public=safe, yellow as private=warning. FWIW this is only really an issue on the first publish when the user forgets to set the proper organisation in mix.exs; afterwards it's hard to make a mistake. So maybe there are some other solutions to achieve the same goal?

josevalim:
`IO.ANSI.enabled?`

wojtekmach:
Ok, I think I was framing the problem wrong. If the problem is that users didn't pay attention where it's being published then it shouldn't matter that much if we differentiate between public/private - it should be enough to have some more emphasis in the first place. And then down the road differentiation is less important because presumably the user configured `mix.exs` correctly.

How about:

1. screen shot 2017-10-30 at 09 22 56

or 

2: screen shot 2017-10-30 at 09 28 59

josevalim:
Here is an idea, why don't we also minimize the amount of information shown?

```
Building package shipit 0.2.3 (use --verbose to see package contents)
All packages must follow Hex Code of Conduct: https://hex.pm/coc
...warnings...

Publishing package to **public** repository hexpm.
Proceed? [Yn]
```

Regarding the Hex Code of Conduct, you could also store a configuration on the user machine with the code of conduct version and show it only the first time the user publishes a package:

```
Before publishing, please read Hex Code of Conduct: https://hex.pm/coc
Agreed? [Yn]
```

Even if you don't agree with those for now, I would at least change the first Publishing mention to Building, otherwise you get Publishing appearing multiple times. :)

josevalim:
@wojtekmach you can also use IO.ANSI.enabled? to alternative between `**public**` and `public` in bold.

wojtekmach:
> I would at least change the first Publishing mention to Building

sounds good to me, applied.

Regarding two other proposals (for CoC and verbose mode) I like them. WDYT @ericmj? Would be a separate PRs obviously.

ericmj:
👍 for changing publishing to building.

I would like to avoid storing more local state because it's fragile and hard to test so I am 👎 on not always showing CoC link.

Not sure about verbose mode because people easily make mistakes, see recent ex_doc issue, and this is one of the only ways for us to show what we are actually packaging. Hiding this output behind an option is effectively removing the output.

josevalim:
> see recent ex_doc issue

Exactly. I don't read those and I don't think anybody does after the package is published for the first time.

ericmj:
> Exactly. I don't read those and I don't think anybody does after the package is published for the first time.

I don't think the solution is to completely remove the output. The only other way we have to verify the package today is the smoke test which is opt-in and does not find all issues that the package contents output can show. Until we have better solution we should keep showing the output.

PR #461, opened 2017-11-25
Add Hex.Shell.debug/1 used with MIX_DEBUG=1

wojtekmach:
Ref https://github.com/hexpm/hex/issues/358#issuecomment-284712035
Ref https://github.com/hexpm/hex/issues/454

without MIX_DEBUG=1:

```
Resolving Hex dependencies...
Dependency resolution completed:
  certifi 1.0.0
  cowboy 1.1.2
  cowlib 1.0.2
  hackney 1.6.6 RETIRED!
    (invalid)
  idna 4.0.0
  metrics 1.0.1
  mime 1.1.0
  mimerl 1.0.2
  ranch 1.3.2
  ssl_verify_fun 1.1.1
* Getting mime (Hex package)
* Getting cowboy (Hex package)
* Getting hackney (Hex package)
* Getting certifi (Hex package)
* Getting idna (Hex package)
* Getting metrics (Hex package)
* Getting mimerl (Hex package)
* Getting ssl_verify_fun (Hex package)
* Getting cowlib (Hex package)
* Getting ranch (Hex package)
```

with MIX_DEBUG=1:

```
** Running mix loadconfig (inside Foo.MixProject)
** Running mix deps.get (inside Foo.MixProject)
** Running mix archive.check (inside Foo.MixProject)
Resolving Hex dependencies...
Dependency resolution completed:
  certifi 1.0.0
  cowboy 1.1.2
  cowlib 1.0.2
  hackney 1.6.6 RETIRED!
    (invalid)
  idna 4.0.0
  metrics 1.0.1
  mime 1.1.0
  mimerl 1.0.2
  ranch 1.3.2
  ssl_verify_fun 1.1.1
* Getting mime (Hex package)
  Using locally cached package (/Users/wojtek/.hex/packages/hexpm/mime-1.1.0.tar)
* Getting cowboy (Hex package)
  Using locally cached package (/Users/wojtek/.hex/packages/hexpm/cowboy-1.1.2.tar)
* Getting hackney (Hex package)
  Using locally cached package (/Users/wojtek/.hex/packages/hexpm/hackney-1.6.6.tar)
* Getting certifi (Hex package)
  Using locally cached package (/Users/wojtek/.hex/packages/hexpm/certifi-1.0.0.tar)
* Getting idna (Hex package)
  Using locally cached package (/Users/wojtek/.hex/packages/hexpm/idna-4.0.0.tar)
* Getting metrics (Hex package)
[...]

PR #462, opened 2017-11-25
[WIP] Include package versions in deps.get

wojtekmach:
This allows us to include version in the output (see https://github.com/hexpm/hex/pull/461 for output without it):

```
Resolving Hex dependencies...
Dependency resolution completed:
  certifi 1.0.0
  cowboy 1.1.2
  cowlib 1.0.2
  hackney 1.6.6 RETIRED!
    (invalid)
  idna 4.0.0
  metrics 1.0.1
  mime 1.1.0
  mimerl 1.0.2
  ranch 1.3.2
  ssl_verify_fun 1.1.1
* Getting mime (Hex package v1.1.0)
* Getting cowboy (Hex package v1.1.2)
* Getting hackney (Hex package v1.6.6)
```

This is the exact output we wanted to have in https://github.com/hexpm/hex/issues/454 (+/- hiding some output when not in debug mode)

However, including the version here makes it a bit redundant in other places:

```
$ mix deps
* idna (Hex package v4.0.0) (rebar3)
  locked at 4.0.0 (idna) 10aaa9f7
  the dependency build is outdated, please run "mix deps.compile"
* mimerl (Hex package v1.0.2) (rebar3)
  locked at 1.0.2 (mimerl) 993f9b0e
  the dependency build is outdated, please run "mix deps.compile"
```

```
$ mix deps.compile && mix deps
* idna 4.0.0 (Hex package v4.0.0) (rebar3)
  locked at 4.0.0 (idna) 10aaa9f7
  ok
* mimerl 1.0.2 (Hex package v1.0.2) (rebar3)
  locked at 1.0.2 (mimerl) 993f9b0e
  ok
```

WDYT?

### DIFF
--- a/lib/hex/scm.ex
+++ b/lib/hex/scm.ex
@@ -11,8 +11,13 @@ defmodule Hex.SCM do
     true
   end
 
-  def format(_opts) do
-    "Hex package"
+  def format(opts) do
+    case Hex.Utils.lock(opts[:lock]) do
+      %{version: version} ->
+        "Hex package v#{version}"
+      nil ->
+        "Hex package"
+    end
   end
 
   def format_lock(opts) do


wojtekmach:
@josevalim are you ok with redundant output too? See the 3rd code snippet in particular, it looks kinda silly because of duplication but I think any other way would require changing Mix.

wojtekmach:
@ericmj I started updating test suite with this change:

```elixir
      assert_received {:mix_shell, :info, ["* ecto 0.2.0 (Hex package v0.2.0)" <> _]}
      assert_received {:mix_shell, :info, ["  locked at 0.2.0 (ecto)" <> _]}
      assert_received {:mix_shell, :info, ["  ok"]}
```

having 0.2.0 in _3_ places really does look kinda silly :) I'm gonna close this for now, we can always add it later.

---

Btw, I briefly looked into adding optional callback `Mix.SCM.format_dep/1` that [Mix.Dep.format_dep/1](https://github.com/elixir-lang/elixir/blob/v1.5.2/lib/mix/lib/mix/dep.ex#L338) could use to have even more control on displaying stuff so that we can _always_ do `foo 0.2.0 (Hex package)` because we always have the lock at that moment so we can always display the version after package name. The only problem was that if dependency name differs from package name we don't have that information in lock, so we'd probably rather need: `format_dep(app, opts)` which starts to getting unnecessarily complicated.

PR #463, opened 2017-11-25
Show all outdated dependencies

obahareth:
Possible fix for #296.

### DIFF
--- a/lib/mix/tasks/hex.outdated.ex
+++ b/lib/mix/tasks/hex.outdated.ex
@@ -114,7 +114,7 @@ defmodule Mix.Tasks.Hex.Outdated do
 
   defp all(deps, lock, opts) do
     values =
-      if(opts[:all], do: deps, else: Enum.filter(deps, & &1.top_level))
+      deps
       |> sort()
       |> get_versions(lock, opts[:pre])
       |> Enum.map(&format_all_row/1)


obahareth:
From what I understood in the discussion mentioned in the issue, the desired solution was to always list all outdated packages without needing to provide a `--all` flag.

obahareth:
Whoops, I was testing using `mix test --exclude integration` (which doesn't make any sense now that I think about it). I'll update the tests and test without excluding the hexpm integration.

wojtekmach:
> the desired solution was to always list all outdated packages without needing to provide a --all flag.

@obahareth not quite. Currently, without `--all` option we only list top-level packages, top-level meaning defined in `mix.exs` of the current project. Deps defined in umbrella child apps are not top-level as far as the root project is concerned (they're considered top-level if the current project is the child app project) and that's why we don't show them. The idea is to output umbrella child's top-level deps.

I'm actually working on a proof-of-concept of this too, stay tuned :)

obahareth:
@wojtekmach  So the goal is to get all top-level deps for the umbrella app itself as well as the top-level deps of child apps when you run `mix hex.outdated`, while `mix hex.outdated --all` will show all outdated dependencies for all apps (including ones that aren't top-level), right?

Should I still update my PR if you're already working on a solution, or should I just close it?

wojtekmach:
@obahareth yes, that's precisely the goal. I'm gonna submit a WIP PR for this in a minute and if you'd like to take a look and/or pick up where I left off that would be very appreciated!

PR #473, opened 2017-12-23
Fix handle_redirect to match calling arguments

matthewoden:
The arguments on handle_redirect will never match (handle_response transforms the three item tuple into a different three item tuple), so the redirect logic will never be called.

I'd love to add a test for this, but the timeout, retry, and redirect methods are all private, and request directly calls httpc. I thought about adding an adapter, but wasn't sure if that'd add too much indirection.


wojtekmach:
Thank you!

PR #475, opened 2018-01-01
Fix grammatical error in documention in hex.retire

supersimple:
This fixes a very small grammatical error in the documentation for the hex.retire mix task.


wojtekmach:
Thank you!

PR #482, opened 2018-01-07
Alerts if a package has been retired in hex.info

supersimple:
prints an alert when info is searched on a package version.
screen shot 2018-01-07 at 11 07 46 am

### DIFF
--- a/lib/mix/tasks/hex.info.ex
+++ b/lib/mix/tasks/hex.info.ex
@@ -159,17 +159,9 @@ defmodule Mix.Tasks.Hex.Info do
   defp print_retirement(nil), do: ""
   defp print_retirement(%{"retirement" => nil}), do: ""
   defp print_retirement(release) do
-    reason = release["retirement"]["reason"]
-    message = release["retirement"]["message"]
-    Hex.Shell.info [[:bright, "This version has been retired "], [:reset, format_retirement_info(reason, message)]]
+    retirement = %{reason: release["retirement"]["reason"], message: release["retirement"]["message"]}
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/hex.info.ex
```diff
@@ -153,6 +156,20 @@ defmodule Mix.Tasks.Hex.Info do
     "~> #{major}.#{minor}.#{patch}#{format_pre(pre)}"
   end
 
+  defp print_retirement(nil), do: ""
+  defp print_retirement(%{"retirement" => nil}), do: ""
+  defp print_retirement(release) do
+    reason = release["retirement"]["reason"]
+    message = release["retirement"]["message"]
+    Hex.Shell.info [[:bright, "This version has been retired "], [:reset, format_retirement_info(reason, message)]]
```
wojtekmach: please use `Hex.Utils.package_retirement_status/1` to get the message, see: https://github.com/hexpm/hex/blob/v0.17.1/lib/mix/tasks/hex.audit.ex#L48
 
maybe let's also use the yellow color as in `hex.audit`?

supersimple: Thanks. Just pushed up those changes.
screen shot 2018-01-07 at 12 49 45 pm

lib/mix/tasks/hex.info.ex
```diff
@@ -153,6 +156,12 @@ defmodule Mix.Tasks.Hex.Info do
     "~> #{major}.#{minor}.#{patch}#{format_pre(pre)}"
   end
 
+  defp print_retirement(nil), do: ""
```
ericmj: We don't need this clause.

wojtekmach:
Thank you!

PR #508, opened 2018-02-16
Make docs project aware

supersimple:
I have made a lot of changes to create the new functionality, but I am having trouble setting up the tests.
If one of the maintainers has a little time, perhaps you can setup an example for me to follow.
One particular issue I have is flipping the value of `def in_mix_project?(), do: !!Mix.Project.get()`

Thanks

### DIFF
--- a/lib/mix/tasks/hex.docs.ex
+++ b/lib/mix/tasks/hex.docs.ex
@@ -64,8 +64,7 @@ defmodule Mix.Tasks.Hex.Docs do
   end
 
   defp fetch_docs([], _opts, true) do
-    deps_in_project()
-    |> Enum.each(fn pkg ->
+    Enum.each(deps_in_project(), fn pkg ->
       fetch_docs([pkg[:name], pkg[:version]], %{organization: pkg[:repo]}, true)
     end)
   end
@@ -157,13 +156,13 @@ defmodule Mix.Tasks.Hex.Docs do
 
   defp open_docs(package, opts, true) do
     package_in_lock =
-      deps_in_project() |> Enum.find(fn pkg -> pkg[:name] == List.first(package) end)
+      Enum.find(deps_in_project(), fn pkg -> pkg[:name] == List.first(package) end)
 
-    if !is_nil(package_in_lock) do
+    if is_nil(package_in_lock) do
+      open_latest_docs(package, opts)
+    else
       version = package_in_lock[:version]
       open_docs([package, version], opts, true)
-    else
-      open_latest_docs(package, opts)
     end
   end
 
@@ -192,15 +191,13 @@ defmodule Mix.Tasks.Hex.Docs do
   end
 
   defp open_docs_offline([name], opts, true) do
-    package_in_lock =
-      deps_in_project()
-      |> Enum.find(fn pkg -> pkg[:name] == name end)
+    package_in_lock = Enum.find(deps_in_project(), fn pkg -> pkg[:name] == name end)
 
-    if !is_nil(package_in_lock) do
+    if is_nil(package_in_lock) do
+      open_latest_docs_offline(name, opts)
+    else
       latest_version = package_in_lock[:version]
       open_docs_offline([name, latest_version], opts, true)
-    else
-      open_latest_docs_offline(name, opts)
     end
   end
 
@@ -283,7 +280,7 @@ defmodule Mix.Tasks.Hex.Docs do
 
   if Mix.env() == :test do
     defp system_cmd(cmd, args) do
-      send self(), {:hex_system_cmd, cmd, args}
+      send(self(), {:hex_system_cmd, cmd, args})
     end
   else
     defp system_cmd(cmd, args) do
--- a/lib/mix/tasks/hex.docs.ex
+++ b/lib/mix/tasks/hex.docs.ex
@@ -41,6 +41,7 @@ defmodule Mix.Tasks.Hex.Docs do
   def run(args) do
     Hex.start()
     {opts, args} = Hex.OptionParser.parse!(args, strict: @switches)
+    opts = convert_organization_to_repo(opts)
     in_mix_project? = !!Mix.Project.get()
 
     case args do
@@ -75,7 +76,7 @@ defmodule Mix.Tasks.Hex.Docs do
 
   defp fetch_docs([], _opts, true = _in_mix_project?) do
     Enum.each(deps_in_project(), fn pkg ->
-      fetch_docs([pkg[:name], pkg[:version]], %{organization: pkg[:repo]}, true, true)
+      fetch_docs([pkg[:name], pkg[:version]], %{repo: pkg[:repo]}, true, true)
     end)
   end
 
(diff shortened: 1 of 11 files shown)


lib/mix/tasks/hex.docs.ex
```diff
@@ -62,34 +63,61 @@ defmodule Mix.Tasks.Hex.Docs do
     end
   end
 
-  defp fetch_docs(_organization, []) do
+  defp fetch_docs([], _opts, true) do
+    deps_in_project()
+    |> Enum.each(fn pkg ->
```
ericmj: Avoid single pipe.

lib/mix/tasks/hex.docs.ex
```diff
@@ -115,35 +143,76 @@ defmodule Mix.Tasks.Hex.Docs do
...
+
+  defp open_docs(package, [latest: true] = opts, _in_mix_project?) do
+    open_latest_docs(package, opts)
+  end
+
+  defp open_docs(package, opts, true) do
+    package_in_lock =
+      deps_in_project() |> Enum.find(fn pkg -> pkg[:name] == List.first(package) end)
+
+    if !is_nil(package_in_lock) do
```
ericmj: Switch the do an else blocks inside of negating.

lib/mix/tasks/hex.docs.ex
```diff
@@ -115,35 +143,76 @@ defmodule Mix.Tasks.Hex.Docs do
...
+    get_docs_url([package, version], opts) |> browser_open()
+  end
+
+  defp open_docs(package, [latest: true] = opts, _in_mix_project?) do
+    open_latest_docs(package, opts)
+  end
+
+  defp open_docs(package, opts, true) do
+    package_in_lock =
+      deps_in_project() |> Enum.find(fn pkg -> pkg[:name] == List.first(package) end)
```
ericmj: No pipe :).

supersimple:
Thanks @ericmj 
I updated the documentation and cleaned up the syntax a bit.

ericmj:
@supersimple I pushed some tests.

wojtekmach:
I tested this on hexpm codebase and noticed a couple of issues:

```
hexpm[master]% mix hex.docs fetch
** (Mix) No package with name ssl_verify_fun or version 1.1.1
hexpm[master]% 
```

1. when we 404 on docs for a particular package we should keep going instead of halting. Some packages don't host their docs on hexdocs.pm, most notably cowboy

2. the error is a bit misleading, I think it should read something like:

   ```
   ** (Mix) Couldn't find docs for package with name ssl_verify_fun or version 1.1.1
   ```

supersimple:
I have pushed a change to that error message, as well as a fix that allows mix hex.docs fetch to continue if a package cannot be found.
screen shot 2018-02-19 at 12 43 26 pm

supersimple:
^ happy to accept feedback/suggestions

wojtekmach:
please rebase! :)

supersimple:
@wojtekmach I will rebase later today.

lib/mix/tasks/hex.docs.ex
```diff
@@ -46,56 +56,83 @@ defmodule Mix.Tasks.Hex.Docs do
...
         mix hex.docs online PACKAGE [VERSION]
         """)
     end
   end
 
-  defp fetch_docs(_organization, []) do
-    Mix.raise("You must specify at least the name of a package")
+  defp fetch_docs([], _opts, true = _in_mix_project?) do
+    Enum.each(deps_in_project(), fn pkg ->
+      fetch_docs([pkg[:name], pkg[:version]], %{organization: pkg[:repo]}, true, true)
```
supersimple: the dependencies seem to use this repo key to determine if the package is private or not (ie. if it is not "hexpm", it is private)
Is there another key available that is organization-specific that I did not see??

ericmj: There maybe is some confusion about repositories, organizations and private packages. Internally to the Hex client there is no concept of private packages. Hex can fetch packages (and docs) from different sources, these sources are the repositories and you can fetch from public repositories other than Hex.pm.

We also have some functionality to work with Hex.pm organizations in the client. Organizations are a feature of Hex.pm that builds on top of repositories, an organization is essentially a private repository with some user permission controls.

In the client an organization with the name "acme" will get the repository name "hexpm:acme" [1]. We need to mirror the same names here so that `docs/` and `tarballs/` get the same directory structure.

[1] https://github.com/hexpm/hex/blob/master/lib/hex/scm.ex#L43-L53

supersimple:
Thanks.
And just to clarify, is there another place in the lock file to derive the organization from?

ericmj:
> And just to clarify, is there another place in the lock file to derive the organization from?

No, you are getting it from the right place. We just need to make sure we use the correct name in urls "https://repo.hex.pm/acme/docs/..." and in directories "$HEX_HOME/docs/hexpm:acme/...".

supersimple:
@wojtekmach @ericmj I _think_ this is code-complete. Is there anything else I need to do before this can be merged?

lib/mix/tasks/hex.docs.ex
wojtekmach: yup, `%{organization: pkg[:repo]}` doesn't seem correct because it's mixing organization with repo, this should be `%{repo: pkg[:repo]}`. This means that `fetch_docs` would need to take a repo and not organization, and a way to handle that is to convert organization into repo when parsing mix task args. Eventually I think we'll support both `--organization ORGANIZATION` and `--repo REPO` (mutually exclusive) in this task and in some other tasks, but we don't have to worry about it yet.

supersimple:
@wojtekmach I changed the internal references to "organization" to "repo" instead. Is that what you were thinking of doing to clarify the distinction?

lib/mix/tasks/hex.docs.ex
```diff
@@ -283,6 +383,18 @@ defmodule Mix.Tasks.Hex.Docs do
...
+    |> Enum.map(fn {_app, info} -> Hex.Utils.lock(info) end)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp org_dir(repo) when repo in [nil, "hexpm"], do: "hexpm"
+  defp org_dir(repo), do: "hexpm:#{repo}"
+
+  defp convert_organization_to_repo(opts) do
+    opts
+    |> Keyword.put(:repo, opts[:organization])
```
wojtekmach: @supersimple this is still not quite right, repo != organization. I'm gonna go ahead and revert the latest commit and merge as is. We can tackle the correct distinction between repo & org in future.

supersimple: ok, thanks. 😅

wojtekmach:
Thank you!

lib/mix/tasks/hex.docs.ex
wojtekmach: oops, somehow the revert didn't get push to branch before merging; I reverted it manually on master in https://github.com/hexpm/hex/commit/9237726e1e2319cc24df23de6f6dfe1427ecc3e3.

PR #535, opened 2018-03-28
Groups deps into unchanged, updated, and downgraded

supersimple:
Here is my initial work on grouping changes. See attached images:
no-changes
one-updated
one-downgraded

A couple of questions:
1. if a group does not have any deps in it, should we make an empty message?
```
Updated:
  No packages updated.
```

2. Should downgraded packages be in yellow?

3. the message from https://github.com/elixir-lang/elixir/blob/abd121d2cd7ade27c03965612f1a7c305d8a1e5e/lib/mix/lib/mix/dep/fetcher.ex#L60
shows if a package is changing. I think that is obsolete with this new formatting. I think the "All dependencies up to date" message makes a little more sense, but even that could be removed in place of a more generic message.

Please let me know what you think and I will be happy to make updates.


supersimple:
meanwhile, let me update those tests...

ericmj:
1) If the group is empty I would just skip it to reduce noise.

2) No, lets keep it green so that retirements stand out, also let's move retirements to it's own group at the bottom. I don't think they make sense to keep under unchanged.

3) I agree, it's a bit confusing that a package can be appear under "Downgraded" and then Mix prints "* Updating ...". Let's change the group name "Updated" to "Upgraded" since updated can mean both directions. I don't think we can remove the Mix message without reducing the experience for other Mix SCMS though.

How we should print when we fetch an unlocked dependency, they don't fit under unchanged/upgraded/downgraded since there was no previous version?

supersimple:
On #3. I have the unlocked deps in with unchanged now, mostly because there is no previous version number to work from. Should there be an additional group for them?

supersimple:
On #2. I'm good with moving retirements to it's own group, but it might confuse things since a package that was updated or downgraded could theoretically be retired.

supersimple:
How about a new group at the top named "Created:" that covers any deps that were not previously locked?

supersimple:
Unchanged with retired
screen shot 2018-03-29 at 10 04 50 am

New package added:
screen shot 2018-03-29 at 10 09 01 am

Includes downgraded:
screen shot 2018-03-29 at 10 09 40 am

supersimple:
> I don't think we can remove the Mix message without reducing the experience for other Mix SCMS though.

can you explain that? what other SCMs could use this bit of code?

ericmj:
> can you explain that? what other SCMs could use this bit of code?

Git dependencies use it.

I think everything looks good. We can go ahead with what we have now, try it out and then tweak it.

lib/hex/remote_converger.ex
```diff
@@ -219,32 +220,93 @@ defmodule Hex.RemoteConverger do
...
+
+        Enum.each(deps, fn {name, repo, previous_version, version} ->
+          print_status(
+            Registry.retired(repo, name, version),
+            mod,
+            name,
+            previous_version,
+            version
+          )
+        end)
```
ericmj: This function is getting a bit big, let's try to break it up into smaller functions.

lib/hex/remote_converger.ex
```diff
@@ -219,32 +220,93 @@ defmodule Hex.RemoteConverger do
...
+
+  defp categorize_dependency_change(nil, _version), do: :new
+
+  defp categorize_dependency_change(previous_version, version) do
+    Version.compare(version, previous_version)
+  end
+
+  defp print_category(mod) do
+    case mod do
+      :new -> Hex.Shell.info("Created:")
```
ericmj: Maybe print "New:" instead of "Created:". WDYT?

supersimple: I'm open, but one instance this happens is when you intentionally unlock a dep first. Dpeneds on how you think of it, but to me that isn't "new"

lib/hex/remote_converger.ex
supersimple: Sure thing. I'll push a refactor shortly

supersimple: Sure thing. I'll push a refactor shortly

lib/hex/remote_converger.ex
supersimple: also, all the other titles end with "ed" now, so it seems like it fits better and is at least as useful/accurate.

supersimple: if you have a strong opinion, feel free to change it. It wont bother me.

wojtekmach: I've changed to "New:" on master. Thanks for the PR!

ericmj:
Thank you! ❤️

paulswartz:
I think this might be causing problems on older lockfiles. I recently updated to 0.17.4 and started seeing the following when installing packages:
```
** (FunctionClauseError) no function clause matching in Hex.RemoteConverger.dep_info_from_lock/1

    The following arguments were given to Hex.RemoteConverger.dep_info_from_lock/1:

        # 1
        {:diff, {:hex, :diff, "1.0.0", "efb2a8422428f8581692f8aa969ef8a546c952039e83e0e27d29005a3772012f", [:mix], []}}

    (hex) lib/hex/remote_converger.ex:261: Hex.RemoteConverger.dep_info_from_lock/1
    (elixir) lib/map.ex:203: Map.new_transform/3
    (hex) lib/hex/remote_converger.ex:224: Hex.RemoteConverger.print_success/3
    (hex) lib/hex/remote_converger.ex:53: Hex.RemoteConverger.converge/2
    (mix) lib/mix/dep/converger.ex:95: Mix.Dep.Converger.all/4
    (mix) lib/mix/dep/converger.ex:51: Mix.Dep.Converger.converge/4
    (mix) lib/mix/tasks/deps.get.ex:27: Mix.Tasks.Deps.Get.run/1
```

ericmj:
Fixed in master, I am doing a new release.

paulswartz:
Thanks so much for the quick response!

wojtekmach:
@supersimple I've updated a couple of projects recently and this feature is really cool, thanks again!

PR #546, opened 2018-04-15
Show more detailed message when fetch times out

supersimple:
This is in response to https://github.com/hexpm/hex/issues/540

I thought we could offer a little more detail in the error message since it seems like it happens fairly regularly to people with poor connectivity. I would like to get feedback on the idea in general and the message itself.

The first commit is very red and bold.
The second commit puts the support message in regular text.

Also, does anyone know how to write a good test for this ?

First commit:
screen shot 2018-04-15 at 10 48 13 am

Second commit:
screen shot 2018-04-15 at 7 43 57 pm

### DIFF
--- a/lib/hex/scm.ex
+++ b/lib/hex/scm.ex
@@ -310,11 +310,12 @@ defmodule Hex.SCM do
           {:error, "Request failed (#{code})"}
 
         {:error, :timeout} ->
-          reason = """
-          Request failed (:timeout)
-          If this happens consistently, adjust your concurrency and timeout settings:
-          `HEX_HTTP_CONCURRENCY=1 HEX_HTTP_TIMEOUT=120 mix deps.get`.
-          """
+          reason = [
+            "Request failed (:timeout)",
+            :reset,
+            "\nIf this happens consistently, adjust your concurrency and timeout settings:",
+            "\n`HEX_HTTP_CONCURRENCY=1 HEX_HTTP_TIMEOUT=120 mix deps.get`."
+          ]
 
           {:error, reason}
 
--- a/lib/hex/scm.ex
+++ b/lib/hex/scm.ex
@@ -309,6 +309,15 @@ defmodule Hex.SCM do
         {:ok, {code, _body, _headers}} ->
           {:error, "Request failed (#{code})"}
 
+        {:error, :timeout} ->
+          reason = """
+          Request failed (:timeout)
+          If this happens consistently, adjust your concurrency and timeout settings:
+          `HEX_HTTP_CONCURRENCY=1 HEX_HTTP_TIMEOUT=120 mix deps.get`.
+          """
+
+          {:error, reason}
+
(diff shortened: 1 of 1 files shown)


lib/hex/scm.ex
```diff
@@ -309,6 +309,16 @@ defmodule Hex.SCM do
         {:ok, {code, _body, _headers}} ->
           {:error, "Request failed (#{code})"}
 
+        {:error, :timeout} ->
+          reason = [
+            "Request failed (:timeout)",
+            :reset,
+            "\nIf this happens consistently, adjust your concurrency and timeout settings:",
+            "\n`HEX_HTTP_CONCURRENCY=1 HEX_HTTP_TIMEOUT=120 mix deps.get`."
```
wojtekmach: how about dropping ` and adding one blank line and 4 chars indent for the command to run?

lib/hex/scm.ex
```diff
@@ -309,6 +309,16 @@ defmodule Hex.SCM do
         {:ok, {code, _body, _headers}} ->
           {:error, "Request failed (#{code})"}
 
+        {:error, :timeout} ->
```
wojtekmach: > Also, does anyone know how to write a good test for this ?

I've manually tested it by adjusting hexpm:

```diff
diff --git a/lib/hexpm/web/controllers/test_controller.ex b/lib/hexpm/web/controllers/test_controller.ex
index 6a121cc9..2c4c436d 100644
--- a/lib/hexpm/web/controllers/test_controller.ex
+++ b/lib/hexpm/web/controllers/test_controller.ex
@@ -44,6 +44,7 @@ defmodule Hexpm.Web.TestController do
   end
 
   def tarball(conn, %{"ball" => ball}) do
+    Process.sleep(:infinity)
     Hexpm.Store.get(nil, :s3_bucket, "tarballs/#{ball}", [])
     |> send_object(conn)
   end
```

and re-running Hex integration tests; I haven't seen the updated error message, so maybe we need to match on something else. Could you double-check?

I'm not sure what's the best way of writing an automated test for this particular case.

lib/hex/scm.ex
supersimple: screen shot 2018-04-16 at 5 50 00 pm

lib/hex/scm.ex
supersimple: screen shot 2018-04-16 at 5 50 15 pm

supersimple: It is quite hard to replicate. You have to have enough connectivity to hit the registry, but then not enough to download the entire package before the timeout. I ran it using `HEX_HTTP_TIMEOUT=2 mix deps.get`

wojtekmach:
@supersimple Thank you!

PS: feel free to write things like `Fixes #546` in PR description so that the issue is automatically closed when PR is merged. Thanks again for your contributions!

PR #553, opened 2018-04-24
Adds new task to generate unencrypted key

supersimple:
Fixes for https://github.com/hexpm/hex/issues/547

Adds a new task, `mix hex.user key --generate` that generates and outputs an unencrypted user key.
Allows including that key in ENV VAR `HEX_API_KEY` when publishing. That will authenticate the user and turn off the confirmation message.

Please give feedback if there is anything that doesnt look right or could be done better.

### DIFF
--- a/lib/hex/state.ex
+++ b/lib/hex/state.ex
@@ -32,7 +32,7 @@ defmodule Hex.State do
         |> default(false)
         |> Kernel.not(),
       clean_pass: true,
-      hex_api_key: load_config(config, ["HEX_API_KEY"], [:"$unencrypted_key"]),
+      hex_api_key: load_config(config, ["HEX_API_KEY"], [:hex_api_key]),
       http_concurrency:
         load_config(config, ["HEX_HTTP_CONCURRENCY"], [:http_concurrency])
         |> to_integer()
--- a/lib/hex/state.ex
+++ b/lib/hex/state.ex
@@ -17,6 +17,7 @@ defmodule Hex.State do
   def init(config) do
     %{
       api_key: load_config(config, [], [:"$encrypted_key", :encrypted_key]),
+      api_key_unencrypted: load_config(config, ["HEX_API_KEY"], [:api_key_unencrypted]),
       api_url:
         load_config(config, ["HEX_API_URL", "HEX_API"], [:api_url])
         |> trim_slash()
@@ -32,7 +33,6 @@ defmodule Hex.State do
         |> default(false)
         |> Kernel.not(),
       clean_pass: true,
-      hex_api_key: load_config(config, ["HEX_API_KEY"], [:hex_api_key]),
       http_concurrency:
         load_config(config, ["HEX_HTTP_CONCURRENCY"], [:http_concurrency])
         |> to_integer()
--- a/lib/hex/state.ex
+++ b/lib/hex/state.ex
@@ -32,6 +32,7 @@ defmodule Hex.State do
         |> default(false)
         |> Kernel.not(),
       clean_pass: true,
+      hex_api_key: load_config(config, ["HEX_API_KEY"], [:"$unencrypted_key"]),
       http_concurrency:
         load_config(config, ["HEX_HTTP_CONCURRENCY"], [:http_concurrency])
         |> to_integer()
--- a/lib/mix/tasks/hex.ex
+++ b/lib/mix/tasks/hex.ex
@@ -76,10 +76,11 @@ defmodule Mix.Tasks.Hex do
                            "Hex requires you to have a local password that applies only to this machine for security " <>
                            "purposes. Please enter it."
 
-  def generate_api_key(username, password, key_name, encrypt? \\ true) do
+  def generate_api_key(username, password, opts) do
     Hex.Shell.info("Generating API key...")
     {:ok, name} = :inet.gethostname()
-    key_name = key_name || List.to_string(name)
+    key_name = opts[:key_name] || List.to_string(name)
+    encrypt? = is_nil(opts[:encrypt]) || opts[:encrypt]
 
     case Hex.API.Key.new(key_name, user: username, pass: password) do
       {:ok, {201, body, _}} ->
@@ -124,7 +125,7 @@ defmodule Mix.Tasks.Hex do
     password = password_get("Account password:") |> Hex.string_trim()
 
     unless opts[:skip_organizations] do
-      case generate_api_key(username, password, opts[:key_name]) do
+      case generate_api_key(username, password, key_name: opts[:key_name]) do
         {:ok, key, password} ->
           Hex.Shell.info("Generating organization keys...")
           auth = [key: key]
--- a/lib/mix/tasks/hex.ex
+++ b/lib/mix/tasks/hex.ex
@@ -87,7 +87,7 @@ defmodule Mix.Tasks.Hex do
           Hex.Shell.info(@local_password_prompt)
           prompt_encrypt_key(body["secret"])
         else
-          body["secret"]
+          Hex.Shell.info(body["secret"])
         end
 
       other ->
--- a/lib/mix/tasks/hex.ex
+++ b/lib/mix/tasks/hex.ex
@@ -76,15 +76,19 @@ defmodule Mix.Tasks.Hex do
                            "Hex requires you to have a local password that applies only to this machine for security " <>
                            "purposes. Please enter it."
 
-  def generate_api_key(username, password, key_name) do
+  def generate_api_key(username, password, key_name, encrypt? \\ true) do
     Hex.Shell.info("Generating API key...")
     {:ok, name} = :inet.gethostname()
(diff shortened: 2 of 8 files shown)


test/mix/tasks/hex.publish_test.exs
```diff
@@ -185,6 +185,34 @@ defmodule Mix.Tasks.Hex.PublishTest do
...
+      Mix.Tasks.Hex.User.run(["key", "--generate"])
+      assert_received {:mix_shell, :info, ["Generating API key..."]}
+      assert_received {:mix_shell, :info, [key]}
+
+      System.put_env("HEX_API_KEY", key)
+      Mix.Tasks.Hex.Publish.run(["package"])
+
+      message =
+        "Package published to http://localhost:4043/packages/release_a/0.0.1 " <>
+          "(003480ccb7e23538f486689bda4f8a5ee50ff495e04a0bfd31e95640b7a6a02a)"
```
wojtekmach: could you change this to:

```
      message = "Building released_name 0.0.1"
      assert_received {:mix_shell, :info, [^message]}
```

as in similar tests? It's less noisy and the checksum aspect is tested once at the top of the module.

lib/mix/tasks/hex.ex
```diff
@@ -76,15 +76,19 @@ defmodule Mix.Tasks.Hex do
                            "Hex requires you to have a local password that applies only to this machine for security " <>
                            "purposes. Please enter it."
 
-  def generate_api_key(username, password, key_name) do
+  def generate_api_key(username, password, key_name, encrypt? \\ true) do
```
ericmj: Let's move the new argument and `key_name` into a new `opts` keyword list instead.

lib/mix/tasks/hex.publish.ex
```diff
@@ -193,7 +193,7 @@ defmodule Mix.Tasks.Hex.Publish do
   end
 
   defp proceed?(build, organization, opts) do
-    confirm? = Keyword.get(opts, :confirm, true)
+    confirm? = Keyword.get(opts, :confirm, show_confirmation_default())
```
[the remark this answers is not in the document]
ericmj: Agree with @wojtekmach.

lib/mix/tasks/hex.publish.ex
```diff
@@ -359,4 +359,16 @@ defmodule Mix.Tasks.Hex.Publish do
...
+
+  defp auth_info() do
+    hex_api_key = System.get_env("HEX_API_KEY")
+
+    if hex_api_key do
+      [key: hex_api_key]
+    else
+      Mix.Tasks.Hex.auth_info()
+    end
+  end
```
ericmj: I'm not sure if we wan't to special case `HEX_API_KEY` in `hex.publish` or make it available for all tasks. I'm leaning towards no special casing, wdyt @wojtekmach?

lib/mix/tasks/hex.user.ex
```diff
@@ -299,4 +310,11 @@ defmodule Mix.Tasks.Hex.User do
         Hex.Utils.print_error_result(other)
     end
   end
+
+  defp generate_unencrypted_key() do
+    # get key from API
+    username = Hex.Shell.prompt("Username:") |> Hex.string_trim()
+    password = Mix.Tasks.Hex.password_get("Account password:") |> Hex.string_trim()
+    Hex.Shell.info(Mix.Tasks.Hex.generate_api_key(username, password, nil, false))
```
ericmj: This `Hex.Shell.info` is probably why @wojtekmach got an error with an invalid password.

lib/mix/tasks/hex.user.ex
```diff
@@ -37,6 +37,12 @@ defmodule Mix.Tasks.Hex.User do
 
     * `--skip-organizations` - Skip deauthorizing all organizations.
 
+  ## Generate API key
+
+  Generates an unencrypted API key for your account
+
+      mix hex.user key --generate
+
```
ericmj: We should also be able to specify the permissions of the API key but we can leave that for later.

lib/mix/tasks/hex.publish.ex
wojtekmach: good point, 👍 for not special casing and making it available in all other tasks.

Please also mention `HEX_API_KEY` in `mix hex.config` docs.

lib/mix/tasks/hex.user.ex
```diff
@@ -117,6 +124,7 @@ defmodule Mix.Tasks.Hex.User do
     mix hex.user whoami
     mix hex.user auth
     mix hex.user deauth
+    mix hex.user key --generate
```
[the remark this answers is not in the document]
supersimple: @wojtekmach what would that extra flag do? And can you explain what happens on the backend to support whatever that is?

supersimple:
@ericmj @wojtekmach I have done most of the work you suggested. Just had one question (see above.)

lib/mix/tasks/hex.ex
supersimple: I think this is moot now that I removed that 4th arg

lib/mix/tasks/hex.user.ex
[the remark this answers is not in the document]
ericmj: You can look at the `mix hex.organization key` task for an example. It names the key so that it's easier to remember why it was generated and how it's used. By default the key name is "HOSTNAME-COUNTER"

lib/mix/tasks/hex.ex
ericmj: 👍

lib/hex/state.ex
```diff
@@ -32,6 +32,7 @@ defmodule Hex.State do
         |> default(false)
         |> Kernel.not(),
       clean_pass: true,
+      hex_api_key: load_config(config, ["HEX_API_KEY"], [:hex_api_key]),
```
ericmj: Let's do `api_key_unencrypted: load_config(config, ["HEX_API_KEY"], [:api_key_unencrypted])` to keep it closer to `:api_key`.

lib/mix/tasks/hex.ex
ericmj: Hold on, I don't see how the `encrypt?` arg is removed? :)

lib/mix/tasks/hex.ex
```diff
@@ -156,11 +160,12 @@ defmodule Mix.Tasks.Hex do
...
+    hex_api_key = Hex.State.fetch!(:hex_api_key)
 
-    if key do
-      [key: prompt_decrypt_key(key)]
-    else
-      authenticate_inline()
+    cond do
+      hex_api_key -> [key: hex_api_key]
+      key -> [key: prompt_decrypt_key(key)]
+      true -> authenticate_inline()
```
ericmj: 👍

lib/mix/tasks/hex.user.ex
```diff
@@ -299,4 +310,11 @@ defmodule Mix.Tasks.Hex.User do
         Hex.Utils.print_error_result(other)
     end
   end
+
+  defp generate_unencrypted_key() do
+    # get key from API
+    username = Hex.Shell.prompt("Username:") |> Hex.string_trim()
+    password = Mix.Tasks.Hex.password_get("Account password:") |> Hex.string_trim()
+    Mix.Tasks.Hex.generate_api_key(username, password, nil, false)
```
ericmj: You would pass the `key_name` here as the third arg.

lib/mix/tasks/hex.user.ex
```diff
@@ -70,12 +81,16 @@ defmodule Mix.Tasks.Hex.User do
       mix hex.user reset_password local
   """
 
+  # TODO: when you revoke the key that is used to set the config api_key_unencrypted, anytime you try to auth after that will fail.
```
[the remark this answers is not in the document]
supersimple: @ericmj There is one issue with the new auth process. If you have an api_key_encrypted setup in your config file, after you revoke all keys (or presumably revoke the key that is used in your config) any operation that uses auth after that will fail until you remove the key from the config.

How do you recommend handling that? maybe a message after running --revoke-all ? not sure about --revoke, we might have to check if you revoked the key you are using in the config.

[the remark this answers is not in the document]
supersimple: Any ideas gentlemen?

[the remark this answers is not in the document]
ericmj: When calling --revoke-all we also run the deauth task. The deauth should probably remove both the encrypted and unencrypted API keys from the config.

For --revoke we can check if the API key used to revoke is the same as being revoked but we cannot check if it's another one stored in the config unless we also store the key name in the config. I think it's fine to not remove the config in this case.

When using a revoked key you should get an error message that it has been revoked so it's not the end of the world if we don't remove it.

ericmj:
Merging this now because I want to rebase my work on organization keys on it.

PR #602, opened 2018-08-18
Updates links in register message

supersimple:
I noticed the hex.user register task emits a message which points to a 404. If you want, I can reduce the number of links here to just include the code of conduct and terms of service.

screen shot 2018-08-18 at 1 23 08 pm


supersimple:
Looks like that page was removed with a redesign somewhat recently:

PR #604, opened 2018-08-19
Micro-optimize http options building

qcam:
Since `proxy_config/1` returns either an empty list or a one-element list, hopefully this commit could make it slightly more efficient.

### DIFF
--- a/lib/hex/http.ex
+++ b/lib/hex/http.ex
@@ -28,12 +28,12 @@ defmodule Hex.HTTP do
   end
 
   defp build_http_opts(url, timeout) do
-    [
+    proxy_config(url) ++ [
       relaxed: true,
       timeout: timeout,
       ssl: Hex.HTTP.SSL.ssl_opts(url),
       autoredirect: false
-    ] ++ proxy_config(url)
+    ]
   end
 
   defp build_request(url, headers, body) do


qcam:
@michalmuskala thanks for the interesting input.

I wrote a small Mix task with Benchee to benchmark it out.

```elixir
defmodule Mix.Tasks.Bench.Test do
  use Mix.Task

  def run(_args) do
    Benchee.run(%{
      "prepend"  => fn ->
        Enum.each(1..10000, fn i ->
          auth_config(i) ++ [b: "b", c: "c", d: "d", e: "e"]
        end)
      end,
      "append"  => fn ->
        Enum.each(1..10000, fn i ->
          [b: "b", c: "c", d: "d", e: "e"] ++ auth_config(i)
        end)
      end
    }, time: 10)
  end

  defp auth_config(i) when i < 50 do
    []
  end

  defp auth_config(i) do
    [proxy: "foo"]
  end
end
```

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i7-6567U CPU @ 3.30GHz
Number of Available Cores: 4
Available memory: 16 GB
Elixir 1.7.2
Erlang 21.0.4
Benchmark suite executing with the following configuration:
warmup: 2 s
time: 10 s
parallel: 1
inputs: none specified
Estimated total run time: 24 s


Benchmarking append...
Benchmarking prepend...

Name              ips        average  deviation         median         99th %
prepend        2.03 K      491.76 μs    ±14.78%         478 μs         780 μs
append         1.73 K      578.03 μs    ±22.13%         566 μs         927 μs

Comparison:
prepend        2.03 K
append         1.73 K - 1.18x slower
```

The result I got on my computer was that `append` is usually ~1.20x slower than `prepend`, could it be me benchmarking it wrongly so it didn't tell the whole truth? Also I haven't done any benchmark in terms of memory used.

This PR is only a random thought I had when browsing the code so feel free to close it if you think it wasn't an improvement as I thought it would.

PR #615, opened 2018-10-16
mix deps.get - deps text color based on status

hlhr:
Fix for Issue #614 

Change the color of the status based on dep-status in deps.get output

New: green
Unchanged: default info color
Downgraded: default warn color (yellow)
Upgraded: green


hlhr:
I checked in the wrong branch, will send the correct one as new PR. Sorry about that.

PR #616, opened 2018-10-16
print color of the status based on dep-status

hlhr:
Fix for Issue #614 

Change the color of the status based on dep-status in deps.get output

New: green
Unchanged: default info color
Downgraded: default warn color (yellow)
Upgraded: green


hlhr:
I checked the failed test and can't reproduce it on my local machine but. Tried it with the same elixir and otp version.

wojtekmach:
Thank you!

PR #636, opened 2018-11-30
Error with a descriptive msg when building a package with git deps

dideler:
Resolves https://github.com/hexpm/hex/issues/635

### DIFF
--- a/lib/mix/tasks/hex.build.ex
+++ b/lib/mix/tasks/hex.build.ex
@@ -248,7 +248,7 @@ defmodule Mix.Tasks.Hex.Build do
         }
       end)
 
-    exclude = Enum.map(exclude, & &1.app)
+    exclude = for {app, _req, _opts} <- exclude, do: app
     {include, exclude}
   end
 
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/hex.build.ex
```diff
@@ -248,7 +248,7 @@ defmodule Mix.Tasks.Hex.Build do
         }
       end)
 
-    exclude = Enum.map(exclude, & &1.app)
+    exclude = for {app, _req, _opts} <- exclude, do: app
```
dideler: I've reverted this line to how it used to be a few years ago: https://github.com/hexpm/hex/commit/9b306180464d4e29a0f6bb822e398875e7ef487f#diff-d958e2ca55c6211c419d1ac9a09cd003R169

The mix task was already supposed to handle non-hex packages as deps, but it couldn't get to that part because it would blow up here due to the missing `:app` property.

Alternatively, we could make `exclude` a list of maps or `Mix.Deps` structs, which seems to be the approach that `include` went with. But I went with the current approach as it introduces less changes.

PR #651, opened 2018-12-30
Update to hex_core v0.4.0

wojtekmach:
- https://github.com/hexpm/hex_core/pull/48
- https://github.com/hexpm/hex_core/pull/51


wojtekmach:
Note to self: use `:hex_registry` instead of calling `:hex_pb_*` directly  before merging (https://github.com/hexpm/hex/blob/master/lib/hex/repo.ex#L288:L289)

PR #654, opened 2019-01-08
Adds organization to config when passing org flag into hex.info

supersimple:
Fixes: #653


wojtekmach:
@supersimple I pushed a commit that should fix the build, although we should streamline test auth flow before this PR lands.

PR #673, opened 2019-02-21
Add saved to location after build

### LINKED ISSUE #664: Add output where the files are being written to `mix hex.build`
Right now we have e.g.:

```
oss/myxql[master]% mix hex.build --unpack
Building myxql 0.1.1
  Dependencies:
    db_connection ~> 2.0 (app: db_connection)
    decimal ~> 1.6 (app: decimal)
    jason ~> 1.0 (app: jason) (optional)
  App: myxql
  Name: myxql
  Files:
    lib
    lib/myxql
    lib/myxql/protocol.ex
    lib/myxql/types.ex
    lib/myxql/auth.ex
    lib/myxql/error.ex
    lib/myxql/query.ex
    lib/myxql/row_value.ex
    lib/myxql/cursor.ex
    lib/myxql/result.ex
    lib/myxql/text_query.ex
    lib/myxql/server_error_codes.ex
    lib/myxql/messages.ex
    lib/myxql.ex
    .formatter.exs
    mix.exs
    README.md
    LICENSE.md
    CHANGELOG.md
  Version: 0.1.1
  Build tools: mix
  Description: MySQL 5.7+ driver for Elixir
  Licenses: Apache 2.0
  Links:
    GitHub: https://github.com/elixir-ecto/myxql
  Elixir: ~> 1.4
```

WDYT about adding a line at the bottom:

```
Saved to ./myxql-0.1.1
```

or something like that? It could be in yellow colour and written to stderr?

supersimple:
Fixes #664

### DIFF
--- a/lib/mix/tasks/hex.build.ex
+++ b/lib/mix/tasks/hex.build.ex
@@ -110,11 +110,13 @@ defmodule Mix.Tasks.Hex.Build do
   defp build_package(meta, output) do
     {_tar, checksum} = Hex.create_tar!(meta, meta.files, output)
     Hex.Shell.info("Package checksum: #{Base.encode16(checksum, case: :lower)}")
+    Hex.Shell.info("Saved to " <> Path.join(["./", output]))
   end
 
   defp build_and_unpack_package(meta, output) do
     {tar, checksum} = Hex.create_tar!(meta, meta.files, :memory)
     %{checksum: ^checksum} = Hex.unpack_tar!({:binary, tar}, output)
+    Hex.Shell.info("Saved to " <> Path.join(["./", output]))
   end
 
   @doc false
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/hex.build.ex
```diff
@@ -110,11 +110,13 @@ defmodule Mix.Tasks.Hex.Build do
   defp build_package(meta, output) do
     {_tar, checksum} = Hex.create_tar!(meta, meta.files, output)
     Hex.Shell.info("Package checksum: #{Base.encode16(checksum, case: :lower)}")
+    Hex.Shell.info("Saved to " <> Path.join(["./", output]))
```
wojtekmach: output can be an absolute path, e.g.: `mix hex.build -o /tmp/a.tar`, in which case this would become `./tmp/a` and be wrong. `-o ../a.tar` would show up as `./../a.tar` which is acceptable.

I think we should drop "./" prefix after all instead of handling all these edge cases (I was the one to propose it 😅), thoughts?

PR #699, opened 2019-06-05
Add --organization option to `mix hex.package`

wojtekmach:


lib/mix/tasks/hex.package.ex
```diff
@@ -128,7 +127,10 @@ defmodule Mix.Tasks.Hex.Package do
         tarball
 
       {:error, reason} ->
-        Mix.raise(reason)
+        Mix.raise(
+          "Downloading " <>
+            Hex.Repo.tarball_url(repo, package, version) <> " failed:\n\n" <> reason
```
wojtekmach: this way the error message is imho a little bit more useful, here's how it looks like:

image

lib/mix/tasks/hex.docs.ex
```diff
@@ -32,7 +32,7 @@ defmodule Mix.Tasks.Hex.Docs do
   ## Command line options
 
     * `--module Some.Module` - Open a specified module documentation page inside desired package
-    * `--organization ORGANIZATION` - The organization the package belongs to
+    * `--organization ORGANIZATION` - Set this for private packages belonging to an organization
```
ericmj: Awesome!

lib/mix/tasks/hex.package.ex
```diff
@@ -66,11 +64,12 @@ defmodule Mix.Tasks.Hex.Package do
   ## Command line options
 
   * `--unpack` - Unpacks the tarball after fetching it
+  * `--organization ORGANIZATION` - The organization the package belongs to
```
ericmj: Also need it here though :)

wojtekmach: oops :D

PR #701, opened 2019-06-06
Verify fetched tarballs

wojtekmach:


lib/hex/registry/server.ex
```diff
@@ -212,7 +213,7 @@ defmodule Hex.Registry.Server do
         :ets.new(@name, [])
 
       {:error, reason} ->
-        Hex.Shell.error("Error opening ETS file #{path}: #{inspect(reason)}")
+        Logger.error("Error opening ETS file #{path}: #{inspect(reason)}")
```
wojtekmach: So this is not ideal. The problem was that `Hex.Shell.error` in tests was sending message to the current process which is this genserver and we were getting a function clause error on handle_info. I'm not sure if using logger here would have unintended consequences besides being kinda ugly:

image

I can think of a couple ways to solve this:

1. Change `Hex.Shell` to allow us to register the process the test shell will send message to
2. Change this GenServer's handle_info to log an error, perhaps only in Mix.env==:test.

any other ideas?

ericmj: I have wanted to fix this for a long time. In all cases except `handle_info` we could return the error message and call `Hex.Shell.error` in the calling process. Would that work in this case?

ericmj:
@wojtekmach I implemented your suggestion with sending the mix shell messages to the test process.

PR #775, opened 2020-03-20
remove reliance on colors for hex.outdated

supersimple:
Closes: #774 
Since Windows does not support the ANSI codes, giving instructions or docs that express meaning through color fails. This might also affect color-blind users. This change add more visual cues.

image
image


supersimple:
LMK what you think of this. Since I have windows now, I am going to try to go thru all of the mix tasks to find other issues.

PR #795, opened 2020-07-22
Make `mix hex.package diff` more CLI-friendly

halostatue:
This makes `mix hex.package diff` more Unix shell friendly. With this, one can do `mix hex.package diff ex_doc 0.22.{0,1}`, which is expanded to `mix hex.package diff ex_doc 0.22.0 0.22.1` and is internally converted to `0.22.0..0.22.1`.

It would be nice to have #705 merged as well, although this PR would require minor updates to be fully compatible with that.


wojtekmach:
This is really nice, thank you! Maybe we should deprecate the “..” variant but so suppose there’s not much harm in keeping it around after all.

PR #843, opened 2021-01-05
Re-fetch package if registry checksum changed

### LINKED ISSUE #821: Task :hex_fetcher erroring with CaseClauseError
* Erlang/OTP: 22
* Elixir: 1.10.1
* Hex: 0.20.6-dev

---

As part of a recent experiment I upgraded my local hex to `0.20.6-dev`. Today, running `mix hex.update oban_web` caused a `CaseClauseError`. Note that `oban_web` is a private hex package, if that makes any difference. Here is a sample of the output, including the stacktrace:

```
Upgraded:
  db_connection 2.2.2 => 2.3.0
  oban_web 2.2.2 => 2.2.3
  postgrex 0.15.6 => 0.15.7

12:08:20.599 [error] Task :hex_fetcher started from #PID<0.445.0> terminating
** (CaseClauseError) no case clause matching: {:ok, <<7, 220, 130, 29, 120, 210, 159, 13, 43, 249, 48, 81, 212, 33, 215, 130, 210, 158, 31, 178, 21, 108, 57, 180, 213, 118, 72, 95, 19, 1, 60, 6>>}
    (hex 0.20.6-dev) lib/hex/scm.ex:400: Hex.SCM.fetch/3
    (elixir 1.10.1) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (elixir 1.10.1) lib/task/supervised.ex:35: Task.Supervised.reply/5
    (stdlib 3.11) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Function: #Function<7.56784114/0 in Hex.SCM.prefetch/1>
    Args: []
* Updating oban_web (Hex package)
```

The stacktrace points to `hex/scm.ex:400`, which was changed in #812 (1ca1f95ccc). I can confirm that reverting to `0.20.5` fixed the issue.

wojtekmach:
Closes #821

I was able to re-produce the above mentioned issue by doing the following:

```
foo$ mix hex.publish package
bar$ mix deps.get
foo$ echo "# foo" >> mix.exs # so that checksum changes
foo$ mix hex.publish package --replace
bar$ mix deps.get
```

The workaround for folks is to remove the cached package: `rm ~/.hex/packages/<repo>/<package>-<version>.tar`.

I'm not sure if doing this rather than failing creates some sort of attack vector, so another option would be to tell folks to manually remove the cached package tarball.

### DIFF
--- a/lib/hex/scm.ex
+++ b/lib/hex/scm.ex
@@ -401,33 +401,51 @@ defmodule Hex.SCM do
         {:ok, ^outer_checksum} ->
           {:ok, :cached}
 
+        {:ok, other_outer_checksum} ->
+          Hex.Shell.warn("""
+          Checksum mismatch between registry and the cached package
+
+          Registry checksum: #{Base.encode16(outer_checksum, case: :lower)}
+          Package checksum:  #{Base.encode16(other_outer_checksum, case: :lower)}
+
+          This may happen when the previously cached package got re-published.
+
+          Re-fetching...\
+          """)
+
+          do_fetch(path, repo, package, version)
+
         {:error, _reason} ->
-          case Hex.Repo.get_tarball(repo, package, version) do
-            {:ok, {200, body, _headers}} ->
-              File.mkdir_p!(Path.dirname(path))
-              File.write!(path, body)
-              {:ok, :new}
-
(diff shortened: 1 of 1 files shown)


ericmj (approved):
💜

PR #849, opened 2021-01-13
Add tests for mix xdg

supersimple:
adds some unit testing to a function added for mix_xdg
I am not unit testing the when the mix_xdg flag is true because it will fail in < OTP19
I also found one other mistake in the last PR with the default arg

### DIFF
--- a/test/hex/config_test.exs
+++ b/test/hex/config_test.exs
@@ -5,7 +5,7 @@ defmodule Hex.ConfigTest do
   test "find_config_home/1 when no env var flags are set" do
     System.delete_env("HEX_HOME")
     {:ok, dir} = Config.find_config_home(:user_data)
-    assert String.match?(dir, ~r/\/.hex/)
+    assert dir =~ ".hex"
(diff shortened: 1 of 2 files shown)


test/hex/config_test.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Hex.ConfigTest do
+  use ExUnit.Case, async: false
+  alias Hex.Config
+
+  test "find_config_home/1 when no env var flags are set" do
+    System.delete_env("HEX_HOME")
+    {:ok, dir} = Config.find_config_home(:user_data)
+    assert String.match?(dir, ~r/\/.hex/)
```
wojtekmach, line 8:
```suggestion
    assert dir =~ ".hex"
```

wojtekmach (commented):
We could add a test for xdg like this:

```
@tag skip: System.otp_release < "19"
test ... do
end
```

Also, lets test for both config and cache dirs since there was a bug?

PR #855, opened 2021-01-18
Gracefully handle missing hex metadata in sponsor task

### LINKED ISSUE #854: hex.sponsor fails, unable to find metadata
```
* (CaseClauseError) no case clause matching: {:error, :enoent}
    (hex 0.21.1) lib/mix/tasks/hex.sponsor.ex:54: anonymous fn/2 in Mix.Tasks.Hex.Sponsor.sponsor_links/2
    (elixir 1.11.2) lib/enum.ex:3421: Enum.flat_map_list/2
    (elixir 1.11.2) lib/enum.ex:3422: Enum.flat_map_list/2
    (hex 0.21.1) lib/mix/tasks/hex.sponsor.ex:40: Mix.Tasks.Hex.Sponsor.run/1
    (mix 1.11.2) lib/mix/task.ex:394: Mix.Task.run_task/3
    (mix 1.11.2) lib/mix/cli.ex:84: Mix.CLI.run_task/2
```

I was about to open up a simple PR to handle this case, but I figured this should not be happening at all, and that there should be no dep that has a missing `hex_metadata.config` file. Indeed, the file is not missing, I assume the path is wrong.

This only ensues if I install hex via `mix local.hex`, if I build the ez archive myself from `main` and install that all is well 🤔

starbelly:
It is possible for a lock file to have entries which no longer exist in a mix project, as such the sponsor task will choke if it finds such an entry and tries to read in a hex_metadata.config file that does not exist.

 - Adjust Mix.Task.Hex.Sponsor.read_metadata/2 (private) to check if the
 path exists before attempt to invoke :file.consult/1 on the file.
- Closes #854

### DIFF
--- a/lib/mix/tasks/hex.sponsor.ex
+++ b/lib/mix/tasks/hex.sponsor.ex
@@ -68,11 +68,7 @@ defmodule Mix.Tasks.Hex.Sponsor do
   end
 
   defp read_metadata(package, deps_path) do
-    path = Path.join([deps_path, package, @metadata_file])
-
-    if File.exists?(path) do
-      :file.consult(path)
-    end
+    :file.consult(Path.join([deps_path, package, @metadata_file]))
   end
 
   defp sponsorship(metadata) do
--- a/lib/mix/tasks/hex.sponsor.ex
+++ b/lib/mix/tasks/hex.sponsor.ex
@@ -60,12 +60,19 @@ defmodule Mix.Tasks.Hex.Sponsor do
             _ ->
               []
           end
+
+        nil ->
+          []
       end
     end)
   end
 
   defp read_metadata(package, deps_path) do
-    :file.consult(Path.join([deps_path, package, @metadata_file]))
+    path = Path.join([deps_path, package, @metadata_file])
+
+    if File.exists?(path) do
+      :file.consult(path)
+    end
   end
 
   defp sponsorship(metadata) do
--- a/lib/mix/tasks/hex.sponsor.ex
+++ b/lib/mix/tasks/hex.sponsor.ex
@@ -61,7 +61,7 @@ defmodule Mix.Tasks.Hex.Sponsor do
               []
           end
 
-        nil ->
+        {:error, :enoent} ->
           []
       end
     end)


ericmj (approved):

lib/mix/tasks/hex.sponsor.ex
```diff
@@ -60,12 +60,19 @@ defmodule Mix.Tasks.Hex.Sponsor do
             _ ->
               []
           end
+
+        nil ->
+          []
```
wojtekmach, line 65:
```suggestion
        {:error, :enoent} ->
          []
```

lib/mix/tasks/hex.sponsor.ex
```diff
@@ -60,12 +60,19 @@ defmodule Mix.Tasks.Hex.Sponsor do
...
     end)
   end
 
   defp read_metadata(package, deps_path) do
-    :file.consult(Path.join([deps_path, package, @metadata_file]))
+    path = Path.join([deps_path, package, @metadata_file])
+
+    if File.exists?(path) do
+      :file.consult(path)
+    end
```
wojtekmach, line 75:
```suggestion
    :file.consult(Path.join([deps_path, package, @metadata_file]))
```

wojtekmach (commented):
Good catch and thanks for instructions how to reproduce it! I was able to see the error by doing the following:

1. add a dependency to mix.exs
2. `mix deps.get`
3. remove the dependency from mix.exs (by default it stays in the mix.lock which is the culprit)
4. run `mix hex.sponsor`

so thanks again!

I left some comments how we can maybe simplify the code a little bit.

Alternatively, and it might be even simpler, we change the function to be:

```elixir
defp read_metadata(package, deps_path) do
  case :file.consult(Path.join([deps_path, package, @metadata_file])) do
    {:ok, metadata} -> metadata
    {:error, :enoent} -> []
  end
end
```

wdyt?

starbelly:
@wojtekmach catching `{:error, :enonent}` makes sense to me 👍 I like the other approach if we're considered about leaking that detail from `read_metadata/2`, but we'd still have to alter the case expression to look out for `[]`, so not much different.

starbelly:
@wojtekmach thanks for adding the reproduction steps ❤️

PR #887, opened 2021-05-24
Adding a tip to run `mix hex.sponsor` if a package with the sponsor link gets downloaded  

kelvinst:
This was previously introduced on #884, got merged to master, but was rolled back because there were failing tests that we didn't realise.

PS.: will work on the failing tests soon.


kelvinst:
Alright, everything should be fixed now @wojtekmach and @ericmj . One important notice though: for when elixir version is < 1.4, the tip is not going to be printed because the `post_converge` callback used to print the message didn't exist back then, is that ok?

ericmj (approved):

wojtekmach:
Thanks, sorry for the delay!

PR #895, opened 2021-07-20
Use Elixir 1.12 for checking formatting

wojtekmach:


ericmj (approved):
Looks like setup-beam v1.8 can no longer install Elixir branches. It tries to install `master` as `vmaster`, so let's revert that change, everything else looks good.

.github/workflows/main.yml
```diff
@@ -5,15 +5,15 @@ on: [push, pull_request]
...
   format:
     name: Format
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
 
       - name: Install OTP and Elixir
-        uses: erlef/setup-beam@v1.7
+        uses: erlef/setup-beam@v1
```
wojtekmach, line 13:
```suggestion
        uses: erlef/setup-beam@v1.7
```

.github/workflows/main.yml
```diff
@@ -82,7 +82,7 @@ jobs:
...
```
wojtekmach, line 85:
```suggestion
        uses: erlef/setup-beam@v1.7
```

PR #980, opened 2023-02-05
Match docs tarball error to package error

jjcarstens:
This helps the cases where the package size may be fine, but the documentation is too large (usually because assets like pictures, gifs, etc have been included). Previously, it would report as `too_big`, but that excludes how big the size is and if it was too big compressed or uncompressed.

This changes the `create_docs/2` error to match the `create/2` error reporting the reason for error and the size that was calculated


jjcarstens:
Just realized this file says it is vendor from `hex_core`, so I moved it over there 🤦🏼 

https://github.com/hexpm/hex_core/pull/138

PR #1007, opened 2023-12-29
Improve logging of auth issues

ream88:
It took me way too long to figure out that the person who set up the private hex.pm registry for me forgot to add billing credentials. I tried everything else because neither the web app nor the command line gave me a proper error. This PR hopefully changes this ;)

### DIFF
--- a/lib/mix/tasks/hex.organization.ex
+++ b/lib/mix/tasks/hex.organization.ex
@@ -277,9 +277,6 @@ defmodule Mix.Tasks.Hex.Organization do
           "Failed to authenticate against organization repository with given key because of: #{message}"
         )
 
-      {:ok, {code, _body, _}} when code in [401, 403] ->
-        Hex.Shell.error("Failed to authenticate against organization repository with given key")
-
       other ->
         Hex.Utils.print_error_result(other)
         Hex.Shell.error("Failed to verify authentication key")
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/hex.organization.ex
```diff
@@ -272,6 +272,11 @@ defmodule Mix.Tasks.Hex.Organization do
       {:ok, {code, _body, _}} when code in 200..299 ->
         :ok
 
+      {:ok, {code, %{"message" => message}, _}} when code in [401, 403] ->
+        Hex.Shell.error(
+          "Failed to authenticate against organization repository with given key because of: #{message}"
+        )
+
       {:ok, {code, _body, _}} when code in [401, 403] ->
         Hex.Shell.error("Failed to authenticate against organization repository with given key")
```
wojtekmach, line 281: Thank you for the PR!

could you check if this clause is still relevant by, say, removing it and seeing if it breaks anything? I don't think we should be getting 401s/403s with or without messages, it should be one or the other ideally.

ream88: Removed it and my local tests are still passing. However the only potential issue I see with removing it, is that my code (`%{"message" => message}`) assumes that the response is JSON and can be parsed into a map with a `message` key.

ream88: So my code could break if the API or for example a load balancer returns a body that is not valid JSON. Question is, if we should try to account for that or not. But that's up to you. :)

wojtekmach: In that case we will fall through to the catch all at the end and we should be good unless I’m missing something. Please remove the redundant clause and I think we are good to go. Thanks!

ream88: Done

wojtekmach:
Thank you!

PR #1008, opened 2024-01-04
Implement exit statuses

ream88:
Thanks for merging 92b48e4279954aeae59a4eaa953e10de059a6e8e, however I just found another small thing we could improve. Currently the mix tasks included in hex do all exit with code `0`, even when an error occurred. This PR changes this to `1` which helps with scripts for example on GitHub CI.

PR #1009, opened 2024-01-04
Set exit status on organization.auth failure

ream88:
See #1008


supersimple (approved):
Looks good. Thanks!

supersimple:
@ericmj 20.3/1.5.3 is continuing to fail, but there is no code here that would require higher versions. And the error from CI is logistic in nature.

PR #1011, opened 2024-01-13
Show warning when an old ex_doc version is used.

jannikbecher:
Screenshot_20240113_135315

This is a feature request from @zachdaniel to remind package maintainers to use the latest ex_doc version.

### DIFF
--- a/lib/mix/tasks/hex.publish.ex
+++ b/lib/mix/tasks/hex.publish.ex
@@ -186,6 +186,7 @@ defmodule Mix.Tasks.Hex.Publish do
     # Elixir v1.15 prunes the loadpaths on compilation and
     # docs will compile code. So we add all original code paths back.
     path = :code.get_path()
+    check_ex_doc_version()
 
     try do
       Mix.Task.run("docs", [])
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/hex.publish.ex
```diff
@@ -326,6 +327,24 @@ defmodule Mix.Tasks.Hex.Publish do
     end
   end
 
+  defp check_ex_doc_version() do
+    case Hex.API.Package.get("hexpm", "ex_doc") do
+      {:ok, {200, body, _headers}} ->
+        current_version = Application.spec(:ex_doc, :vsn) |> List.to_string()
```
jannikbecher, line 333: is there a better way to check for the used ex_doc version?

wojtekmach:
My understanding was Jose suggested to check version when generating the docs and not when publishing them. That is, check in ExDoc and not in Hex.

Could you send a PR to ExDoc instead?

Credo used to and perhaps still has a similar version check.

A more efficient way to test is :hex_repo.get_package, https://hexdocs.pm/hex_core/hex_repo.html#get_package/2.

I’m not sure we should add hex_core to ExDoc, I’d like to keep deps to minimum. We can start there and see what the feedback is. We can always cheat and use hex core that is vendored with Hex, :mix_hex_repo.

PR #1022, opened 2024-05-04
Handle empty package name in mix hex.info

sabiwara:
Currently, it fails as follows:

```sh
$ mix hex.info ""
** (ArgumentError) the Access module supports only keyword lists (with atom keys), got: "meta"

If you want to search lists of tuples, use List.keyfind/3
    (elixir 1.16.2) lib/access.ex:345: Access.get/3
    (hex 2.0.6) lib/mix/tasks/hex.info.ex:108: Mix.Tasks.Hex.Info.print_package/2
```

This is because we call GET packages/ on hexpm which returns a list of objects.

The second commit adds some guards on the API layer to calling the wrong URLs in the first place.

Note: While the `mix hex.info ""` command might not make sense as user input, it might happen when scripting.

### DIFF
--- a/lib/hex/api/package.ex
+++ b/lib/hex/api/package.ex
@@ -3,7 +3,7 @@ defmodule Hex.API.Package do
 
   alias Hex.API
 
-  def get(repo, name, auth \\ []) do
+  def get(repo, name, auth \\ []) when name != "" do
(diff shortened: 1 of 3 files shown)


ericmj (approved):
Thank you!

Issue #1054, opened 2024-11-17
CI env variable displays warning if it's a string


altdsoy:
Hello,

I'm not sure if the description is good enough, but I'm using a CI tool (Woodpecker CI) that injects the `CI` ENV variable with a string value.

It seems that when hex is checking its configuration it also checks this ENV.

In the following commit: https://github.com/hexpm/hex/commit/5eb05bdeb1e2632c67ff1ae44448a0c38617cf7a , its value is checked against a boolean:

```elixir
def to_boolean(nil), do: {:ok, nil}
def to_boolean(false), do: {:ok, false}
def to_boolean(true), do: {:ok, true}
def to_boolean("0"), do: {:ok, false}
def to_boolean("1"), do: {:ok, true}
def to_boolean("false"), do: {:ok, false}
def to_boolean("true"), do: {:ok, true}
def to_boolean("FALSE"), do: {:ok, false}
def to_boolean("TRUE"), do: {:ok, true}
def to_boolean(_), do: :error
```

In my case because I don't have control on this ENV variable, Hex ends up considering it's an error and  I'm getting an annoying warning on every `mix` command in my CI pipeline:

`Invalid Hex config, falling back to default. Source: environment variable CI= "woodpecker"`.

I would like to open a PR but I wanted to discuss for an idiomatic way to solve this.

We can simply assume a truthy value for this ENV in particular (maybe using something like `to_truthy_boolean`) ?

supersimple:
That does sound annoying.
It is also an interesting behavior by woodpecker to not allow their predefined env vars to be modified.
Admittedly, this could just as easily my classified as a design error in their product as a deficiency in the Hex API.
If we were to make a change, my suggestion would be to write a special function like [this](https://github.com/hexpm/hex/blob/f963a8f6ef7a26774d6f97a93d78f8fe7e0a3dea/lib/hex/state.ex#L339) to deal with the CI value (and treat it as you said, truthy or non-truthy. But, keeping that logic hidden inside of a private function.)
If you put up a PR, I am compelled to approve it. @ericmj @wojtekmach any objections?

wojtekmach:
Are you able to overwrite that value eg `CI=true mix deps.get` or set

```yaml
environment:
  CI: true
```

Or something along those lines?

But if not, your patch is fine by me, I think it’d be nice on principle to continue being stricter but in this particular case I can’t see how it could be harmful.

jozuas:
For anyone reading this:

```
environment:
  CI: true
[...]

PR #1055, opened 2024-11-17
handle truthy boolean for CI env variable

altdsoy:
Hello, this is following this issue #1054 

While this issue is not really important (just annoying), here I'm suggesting a fix.
I tried to keep what's being done and just leverage the existing `to_boolean`.

I'm not sure about the naming though (maybe `to_boolean_truthy` is better-ish?).

Feel free to comment...

Thanks!


supersimple:
Thanks @altdsoy 🥇

PR #1058, opened 2025-01-13
Support :warn_if_outdated in deps

wojtekmach:
Example:

    $ mix deps.get
    Resolving Hex dependencies...
    Resolution completed in 0.0s
    Unchanged:
      ecto 3.3.1
      ecto_sql 3.3.2
      ex_doc 0.1.0
      postgrex 0.2.1
    warning: the following deps set `warn_if_outdated: true` and are outdated:

     * ecto 3.3.2 is available
     * ecto_sql 3.3.3 is available

### DIFF
--- a/lib/hex/remote_converger.ex
+++ b/lib/hex/remote_converger.ex
@@ -152,15 +152,13 @@ defmodule Hex.RemoteConverger do
 
         versions =
           versions
-          |> Enum.filter(&Version.match?(&1, requirement))
+          |> Enum.filter(fn vsn -> vsn.pre == [] and Version.match?(vsn, requirement) end)
           |> Enum.sort(&(Version.compare(&1, &2) == :gt))
 
-        latest_version = Enum.find(versions, &(&1.pre == [])) || List.first(versions)
-
-        {:hex, _name, version, _chhecksum, _managers, _, ^repo, _checksum} =
-          Map.fetch!(new_lock, String.to_atom(name))
-
-        if Version.compare(latest_version, version) == :gt do
+        with [latest_version | _] <- versions,
+             {:hex, _name, version, _checksum, _managers, _, ^repo, _checksum} <-
+               Map.fetch!(new_lock, String.to_atom(name)),
+             :gt <- Version.compare(latest_version, version) do
           {name, latest_version}
         end
       end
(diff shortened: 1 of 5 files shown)


lib/hex/remote_converger.ex
```diff
@@ -144,6 +144,39 @@ defmodule Hex.RemoteConverger do
...
+        if Version.compare(latest_version, version) == :gt do
+          {name, latest_version}
+        end
+      end
+      |> Enum.filter(& &1)
+
+    if deps_to_warn != [] do
+      IO.warn(
+        [
+          "the following deps set `warn_if_outdated: true` and are outdated:\n\n",
```
josevalim, line 171: We don't use backticks on Elixir but maybe we do on Hex. Please double check. :)

wojtekmach:
In terms of UX, and this may be a bit tricky finding exact line, but for completeness something like this should be possible most of the time too:

image

lib/hex/remote_converger.ex
```diff
@@ -144,6 +144,39 @@ defmodule Hex.RemoteConverger do
...
     Hex.SCM.prefetch(new_lock)
+
+    deps_to_warn =
+      for %{repo: repo, name: name, requirement: requirement, warn_if_outdated: true} <- requests do
+        {:ok, requirement} = Version.parse_requirement(requirement)
+        {:ok, versions} = Registry.versions(repo, name)
+
+        latest_version =
+          versions
+          |> Enum.filter(&Version.match?(&1, requirement))
```
ericmj, line 155: Lets filter pre-releases also

ericmj (approved):

lib/hex/remote_converger.ex
```diff
@@ -144,6 +144,40 @@ defmodule Hex.RemoteConverger do
...
+          |> Enum.sort(&(Version.compare(&1, &2) == :gt))
+
+        latest_version = Enum.find(versions, &(&1.pre == [])) || List.first(versions)
+
+        {:hex, _name, version, _chhecksum, _managers, _, ^repo, _checksum} =
+          Map.fetch!(new_lock, String.to_atom(name))
+
+        if Version.compare(latest_version, version) == :gt do
+          {name, latest_version}
+        end
```
josevalim, line 165:
```suggestion
          |> Enum.filter(fn vsn -> vsn.pre == [] and Version.match?(vsn, requirement) end)
          |> Enum.sort(&(Version.compare(&1, &2) == :gt))

        with [latest_version | _] <- versions,
             {:hex, _name, version, _checksum, _managers, _, ^repo, _checksum} <-
               Map.fetch!(new_lock, String.to_atom(name)),
             :gt <- Version.compare(latest_version, version) do
          {name, latest_version}
        end
```

josevalim (approved):

PR #1061, opened 2025-01-23
Support json output

TylerWitt:
This supports passing a `--json` flag to `mix hex.outdated` for programmatic use.

In order to make the change a bit easier, I separated out 3 things. Previously, the task was laid out as a `single` vs `all` format, where both versions dealt with similar logic.

I split them in to three pieces.
* `process_lockfile` -> This basically formats all the dependencies into a tuple `{package, lock_version, latest_version, requirements, outdated?}`
* `display_outdated` -> This structures the view layer--either as a table (existing logic), or as json.
* `set_exit_code` -> This sets the exit code based on whether things are outdated, and whether or not certain opts have been passed.


ericmj:
Sorry for the delay. 
Thank you! 💜

PR #1073, opened 2025-05-09
Fix Elixir versions order in publish script

wojtekmach:
Before this patch, `mix local.hex` would pick Hex precompiled against oldest Elixir version:

```
HEX_BUILDS_URL=http://localhost:8000 mix local.hex --force && mix hex.info
* creating .mix/archives/hex-2.2.1-otp-26
Hex:    2.2.2-dev
Elixir: 1.19.0-dev
OTP:    26.2.5.11

Built with: Elixir 1.17.3 and OTP 26.2.5.11
```

now it will pick the newest:

```
HEX_BUILDS_URL=http://localhost:8000 mix local.hex --force && mix hex.info
* creating .mix/archives/hex-2.2.1-otp-26
Hex:    2.2.2-dev
Elixir: 1.19.0-dev
OTP:    26.2.5.11

Built with: Elixir 1.18.0 and OTP 26.2.5.6
```

In other words, the new hex.csv now matches the order of legacy hex-1.x.csv.

### DIFF
--- a/scripts/release_hex.sh
+++ b/scripts/release_hex.sh
@@ -26,15 +26,17 @@ function main {
   touch "${hex_csv}"
   sed -i.bak "/^${hex_version},/d" "${hex_csv}"
 
-  # UPDATE THIS FOR EVERY RELEASE
+  # UPDATE THIS FOR EVERY RELEASE, ORDER MATTERS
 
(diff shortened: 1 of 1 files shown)


wojtekmach:
I don't think we need to update hex.csv right now, I don't think it hurts to use the older version.

josevalim (approved):

PR #1075, opened 2025-06-02
Add client identifier header to HTTP requests

sorentwo:
Creates an anonymized repository identifier based on the SHA of the first commit, hashed with SHA256 for additional privacy. The identifier is sent as an `x-hex-client-id` header when available.

For example, in the `hex` repo itself:

```
iex(1)> Hex.Utils.client_identifier()
"0f3d954cda05af53f8921bc77be8d2fe1697daded03115fd17b7a3aef32b23b1"
```

The value can then be used by hex.pm and private hex servers to more accurately count requests. That will give a clearer picture of how many unique applications are using a particular package, rather than just the raw number of downloads.

---

Some followup steps if this approach is accepted:

- [ ] Fix test failures in CI from environment mismatch
- [ ] Potentially cache the value to prevent repeated system calls
- [ ] Add a flag to disable sending the client id

/cc @wojtekmach


lib/hex/utils.ex
```diff
@@ -324,4 +324,24 @@ defmodule Hex.Utils do
...
+  anonymization.
+
+  Returns `nil` if git is not available or the directory is not a git repository.
+  """
+  def client_identifier do
+    with path when is_binary(path) <- System.find_executable("git"),
+         {output, 0} <- System.cmd("git", ["rev-list", "--max-parents=0", "HEAD"]) do
+      output
+      |> String.trim()
+      |> then(&:crypto.hash(:sha256, &1))
```
RoyalIcing, line 341: Since cryptographic hashing is deterministic, does this actually get closer to anonymization?

sorentwo: The source is an irreversible, effectively random, SHA from the repository. Hashing that hash shuffles that random data one more time, mostly to indicate that we truly don't want any information about the application or its source.

As it's irreversible start, this may not make it _more_ anonymous, but I don't believe it hurts.

lib/hex/http.ex
```diff
@@ -271,6 +273,13 @@ defmodule Hex.HTTP do
     host
   end
 
+  defp add_client_identifier_header(headers) do
+    case Hex.Utils.client_identifier() do
+      nil -> headers
+      identifier -> Map.put(headers, "x-hex-client-id", identifier)
```
ericmj, line 279: What about x-hex-repo-id?

lib/hex/utils.ex
```diff
@@ -324,4 +324,36 @@ defmodule Hex.Utils do
...
+  anonymization.
+
+  Returns `nil` when:
+
+  - The `HEX_REPO_IDENTIFIER` environment variable is set to anything other `1` or `true`
+  - The `git` executable isn't available 
+  - The current directory isn't within a git repository
+  """
+  def repo_identifier do
+    with :unset <- Process.get(:hex_repo_identifier, :unset),
```
sorentwo, line 341: It's not clear how useful this will be in practice, as I imagine fetching deps is parallelized across multiple processes. However, the codebase appears to support back to Elixir v1.6, hence OTP 19, which is before `persistent_term` was available and we can't cache between processes very easily.

What does everybody think? Is this useful? Is it worth looking at ETS for caching?

sorentwo: Thanks for the tip. I switched to `Hex.State` to check whether the identifier is enabled or disabled, as well as caching the value, which is far more convenient.

Do you think they should be combined into a single `repo_identifier` value that's either `nil`, `false`, or the cached binary?

lib/hex/utils.ex
```diff
@@ -324,4 +324,42 @@ defmodule Hex.Utils do
...
+  anonymization.
+
+  Returns `nil` when:
+
+  - The `HEX_REPO_IDENTIFIER` environment variable is set to anything other `1` or `true`
+  - The `git` executable isn't available 
+  - The current directory isn't within a git repository
+  """
+  def repo_identifier do
+    case Hex.State.get(:repo_identifier) do
```
ericmj, line 341: This may need to be a lock instead since multiple concurrent HTTP requests will race this function.

lib/hex/utils.ex
```diff
@@ -324,4 +324,42 @@ defmodule Hex.Utils do
...
+      cached when is_binary(cached) ->
+        cached
+
+      nil ->
+        cmd_args = ~w(rev-list --max-parents=0 HEAD)
+
+        with false <- Hex.State.get(:no_repo_identifier),
+             path when is_binary(path) <- System.find_executable("git"),
+             {output, 0} <- System.cmd("git", cmd_args, stderr_to_stdout: true) do
```
ericmj, line 350: If this returns non-zero, let's print an error message with `Hex.Shell.debug`.

ericmj (commented):
Is this considered PII and how do we need to inform users about this change other than a note in the changelog?

Other tracking we've talked about adding is:
* Session ID header, which would be a unique ID for each `mix deps.get`
* A header if the env var `CI` is set, to help identity real users vs CI runs

sorentwo:
@ericmj Took me a while, but I've addressed both of your requests.

ericmj (approved):

lib/hex/repo_identifier.ex
```diff
@@ -0,0 +1,72 @@
+defmodule Hex.RepoIdentifier do
+  @moduledoc """
+  Gets an anonymized identifier for the current git repository.
+
+  This module caches the SHA of the first commit in the repository and hashes it once more for
+  anonymization.
+
+  Returns `nil` when:
+
+  - The `HEX_REPO_IDENTIFIER` environment variable is set to anything other `1` or `true`
```
hugobarauna, line 10: Maybe it should be “ HEX_NO_REPO_IDENTIFIER”

?

hugobarauna: I mean, the doc here says: `HEX_REPO_IDENTIFIER`

But the code is actually using `NO_HEX_REPO_IDENTIFIER`:

https://github.com/sorentwo/hex/blob/5e329783583aa4d5929953ce95a73d68590406d2/lib/hex/state.ex#L115-L116

So maybe the docs should be changed to `NO_HEX_REPO_IDENTIFIER`.

Or, I'm missing something. 😅

ericmj:
We need to ask for consent before sending this telemetry according to GDPR because it creates a persistent identifier and is not required for the functionality of Hex.

Since it has limited usefulness when it's opt-in I am inclined to remove it.

cc @sorentwo

hugobarauna:
> We need to ask for consent before sending this telemetry according to GDPR because it creates a persistent identifier and is not required for the functionality of Hex.
> 
> Since it has limited usefulness when it's opt-in I am inclined to remove it.
> 
> cc @sorentwo

@ericmj I’m not sure this identifier would fall under GDPR.

It’s a git repo–scoped value, not tied to any user or machine. Different people or machines using the same repo would generate the same identifier, reinforcing that it’s linked to the project/git repo, not to a person.

I guess GDPR only applies to data that can identify a natural person, not to identifiers of software artifacts like a Git repository. If it were a machine-specific identifier (like a device ID, IP, or cookie), then it could make a person indirectly identifiable — that’s when GDPR applies. But this one isn’t, so it should be outside GDPR’s scope.

But I'm not an expert on GDPR. Only saying that based on my research and personal experience.

ericmj:
@hugobarauna I see your point but it's enough of an edge case where I think the risk is not worth it.

In most cases a repository is not identifiable to a single person but I could imagine cases where it could be argued that a repository identifier is tied to a single person. For example if you start working on a new project for a month before publishing it to github under your username, any `mix deps.get` HTTP requests for that repository before you published it could be attributed to you with reasonable confidence.

PR #1076, opened 2025-07-17
WIP: mix hex.docs.search task

zachdaniel:
Hey folks! This is an initial foray into defininga mix task for searching hex docs. It does the *bare minimum* right now, but gives us a starting point for discussion. Things to discuss likely include:

1. Should this be a thing at all?
2. Should we present human readable output? If so, how do you want to go about handling JSON, as Jason isn't a part of Hex and not all Elixir versions support the new JSON module etc.
3. how should we handle errors? Should we retry?

This task would be useful for agentic tools in addition to just human use, assuming that we format it as a human readable output. Right now it just dumps the raw JSON (which is fine for agents but bad for humans).

Open to thoughts/critiques, or happy to have this PR closed if this doesn't seem like something we want, or isn't in the place we want it.

### DIFF
--- a/lib/mix/tasks/hex.docs.search.ex
+++ b/lib/mix/tasks/hex.docs.search.ex
@@ -1,0 +1,130 @@
+defmodule Mix.Tasks.Hex.Docs.Search do
+  use Mix.Task
+
+  @shortdoc "Searches hexdocs, returning JSON"
+
+  @moduledoc """
+  Searches hexdocs, returning JSON.
+
+  If no version is specified, defaults to version used in the current mix project.
+  If called outside of a mix project or the dependency is not used in the
+  current mix project, defaults to the latest version.
+
+  ## Search documentation for all dependencies in the current mix project
+
+      $ mix hex.docs.search "search term"
+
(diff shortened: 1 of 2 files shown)


zachdaniel:
I've fleshed this out very nicely in `usage_rules`, and I'm very happy to move some or all of that logic here ❤️ 

I wanted a good way to distribute and work on this in a way taht doesn't rely on users updating hex, and I wanted to be able to play w/ packages and formats etc.


https://github.com/user-attachments/assets/d946886a-631c-4e74-bd9d-a3d257ae458f

lib/mix/tasks/hex.docs.search.ex
```diff
@@ -0,0 +1,130 @@
+defmodule Mix.Tasks.Hex.Docs.Search do
```
ericmj, line 1: Would this task fit better under the existing `mix hex.seach` task? Less pollution of the task namespace.

ericmj:
Closing in favor of https://github.com/hexpm/hex/pull/1163

PR #1077, opened 2025-07-17
Validates that a hex package does not include a dpendncy that is in a…

supersimple:
…n umbrella.

...I will grab a screen capture of my local env to show the behavior...


supersimple:
screenshot

ericmj (approved):

PR #1081, opened 2025-08-18
Add :only column and filtering to mix hex.outdated

rossvz:
I'm curious if we'd be open to extending the `mix hex.outdated` command to show the type of dependency - eg prod/test/dev

Especially in larger codebases dependency updates for test-only or dev-only dependencies may flow through a different process and easier to see impact.

## changes
- adds a new `type` column to `mix hex.outdated` that shows the type of dependency
- adds new `--dev` and `--test` flags to filter the list of dependencies:

Screenshot 2025-08-18 at 4 30 35 PM
Screenshot 2025-08-18 at 4 30 30 PM


rossvz:
@ericmj I like your idea, I wasnt aware `only` could use any string, TIL! This is now much more flexible.

changes are implemented in https://github.com/hexpm/hex/pull/1081/commits/664f8117e82c459111030df9e991e223246e8fec

Screenshot 2025-08-22 at 8 06 35 AM


Screenshot 2025-08-22 at 8 06 44 AM

rossvz:
> @rossvz could you look into implementing this suggestion (emphasis mine)?
> 
> 
> 
> > Name the column "Only" instead of "Type", **if :only is not set then leave the column blank**

@wojtekmach This has been implemented in the latest commit (see screenshots above as well)

wojtekmach:
In the first of two screenshots the only flag is NOT passed however the only column is shown in the output.

lib/mix/tasks/hex.outdated.ex
```diff
@@ -225,6 +233,26 @@ defmodule Mix.Tasks.Hex.Outdated do
...
+        # deps can have multiple `only` values, so we separate by `,`
+        only_values = String.split(only_value, ",", trim: true)
+
+        Enum.filter(versions, fn [_package, dep_only, _lock, _latest, _requirements] ->
+          if String.contains?(dep_only, ",") do
+            dep_only_parts = String.split(dep_only, ",")
+            Enum.any?(dep_only_parts, &(&1 in only_values))
+          else
+            dep_only in only_values
+          end
```
ericmj, line 251:
```suggestion
          dep_only_parts = String.split(dep_only, ",")
          Enum.any?(dep_only_parts, &(&1 in only_values))
```

lib/mix/tasks/hex.outdated.ex
```diff
@@ -329,12 +360,45 @@ defmodule Mix.Tasks.Hex.Outdated do
...
+      nil ->
+        ""
+
+      dep_configs ->
+        # Get the opts from the first (main project) dependency config
+        case List.first(dep_configs) do
+          {_src, _req, opts} -> dep_only_from_opts(opts)
+          _ -> ""
+        end
+    end
```
ericmj, line 383:
```suggestion
    # Get the opts from the first (main project) dependency config
    dep_configs = Map.get(deps, dep_name, [])

    case List.first(dep_configs) do
      {_src, _req, opts} -> dep_only_from_opts(opts)
      _ -> ""
    end
```

lib/mix/tasks/hex.outdated.ex
```diff
@@ -329,12 +360,45 @@ defmodule Mix.Tasks.Hex.Outdated do
...
+        Atom.to_string(only_value)
+
+      only_value when is_list(only_value) ->
+        # For multiple environments, we'll show them comma-separated in the display
+        only_value
+        |> Enum.map(&Atom.to_string/1)
+        |> Enum.join(",")
+
+      only_value ->
+        to_string(only_value)
```
ericmj, line 401:
```suggestion
    Keyword.get(opts, :only, [])
    |> List.wrap()
    |> Enum.map(&to_string/1)
    |> Enum.join(",")
```

ericmj (commented):
Here are some suggested changes. Also please take a look at the CI failures.

rossvz:
@ericmj Thanks for the suggestions! I've tackled the ones you mentioned and attempted to fix failing tests. I am getting a few failures locally that are probably related to test config but not the changes made here (as far as I can tell)

ericmj (approved):

PR #1101, opened 2025-10-27
Improve deps.get error messages for 404 and 403 status codes

varjas:
Improve error messages for certain scenarios described in: https://github.com/hexpm/hex/issues/627

Please let me know if there's any suggestions to adjust the error message text, or if the print logic should be moved to a separate module with a public interface instead of using the `@doc false` syntax to facilitate testing.

### DIFF
--- a/lib/hex/registry/server.ex
+++ b/lib/hex/registry/server.ex
@@ -440,7 +440,8 @@ defmodule Hex.Registry.Server do
     end
   end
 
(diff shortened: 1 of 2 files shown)


ericmj:
Thanks for sending this PR, it helped shaping auth error improvements when we added OAuth support. I think this PR now largely duplicates functionality we already have in Hex.

Thank you! 💜

PR #1103, opened 2025-10-28
Include stdlib by default in mix hex.search

josevalim:


lib/mix/tasks/hex.search.ex
```diff
@@ -7,9 +7,11 @@ defmodule Mix.Tasks.Hex.Search do
   Open and perform searches.
 
   When invoked without arguments, it opens up a search page on
-  https://hexdocs.pm with all of your dependencies selected:
+  https://hexdocs.pm with all of your dependencies plus standard
```
wojtekmach:
```suggestion
  https://hexdocs.pm with all of your dependencies plus Elixir standard
```

lib/mix/tasks/hex.search.ex
```diff
@@ -7,9 +7,11 @@ defmodule Mix.Tasks.Hex.Search do
...
+  library selected:
 
       $ mix hex.search
+      $ mix hex.search --no-stdlib
```
wojtekmach: curious, if we'd ever opt-in/opt-out to OTP, would it be `--otp`/`--no-otp`? There's name overlap between this _Elixir_ stdlib and OTP's stdlib app but I don't think it matters. Another idea is `--no-elixir` but that's probably worse. All that to say, this looks good to me as is, thought I'd add some things that crossed my mind.

wojtekmach (approved):

PR #1125, opened 2026-03-21
Fix registry ets package timestamp purging

elfenlaid:
The changes fix typo in `timestamp` and now `Hex.Registry.Server.prefetch` should properly clean all entries associated with packages from `state.ets`.

The rest of the file uses `timestamp` keys, like here: https://github.com/hexpm/hex/blob/8886724c138c70fa049eaa0566e3219a1ac191eb/lib/hex/registry/server.ex#L368

I also checked for typos in other keys or key that could be omitted from `purge_repo_matchspec`, just in case, but everything seems fine aside from that typo in `timestamp`.


ericmj:
Nice find. Thank you! 💜

PR #1137, opened 2026-04-10
Show mix hex.package diff commands in hex.outdated output

ericmj:
The goal is to provide informations to agents on how to use view package diffs as they cannot easily access https://diff.hex.pm.

image


ericmj:
It adds quite a but of output. Should we only show a single example command instead of a command for every outdated package? Or maybe add it to the table, but I am worried about the width of the table.

wojtekmach:
Hard to decide, I'm too worried about the width of the table so that's probably a -1 for that. I think showing just one command, or a generic one like, say, `mix hex.package diff {package} {from}..{to}` is a good compromise. Another option is to have an `--` flag for more output but discoverability.

PR #1143, opened 2026-04-13
Update to use `hex_auth` module

maennchen:
Upstream PR: https://github.com/hexpm/hex_core/pull/178


lib/hex/api/oauth.ex
```diff
@@ -66,36 +66,22 @@ defmodule Hex.API.OAuth do
...
-      if api_url do
-        Map.put(config, :api_url, api_url)
-      else
-        config
-      end
-
-    scope_string = if is_list(scopes), do: Enum.join(scopes, " "), else: scopes
-    opts = if name, do: [name: name], else: []
-    :mix_hex_api_oauth.client_credentials_token(config, @client_id, api_key, scope_string, opts)
+    :mix_hex_api_oauth.device_auth_flow(config, @client_id, scopes, prompt_user, opts)
```
ericmj: `:mix_hex_api_oauth.device_auth_flow()` returns atoms keys, but most places expect string keys, such as here: https://github.com/hexpm/hex/pull/1143/changes#diff-32a2f809701e24e828ce1c5515cdb72a1a1948918391cadf707a60a328706af7R69

lib/hex/auth.ex
```diff
@@ -0,0 +1,145 @@
...
+      nil ->
+        case Hex.Repo.fetch_repo(repo) do
+          {:ok, config} -> config
+          :error -> :undefined
+        end
+
+      api_key ->
+        %{api_key: api_key}
+    end
+  end
```
ericmj, line 56: This looks strange to me, we first attempt to get API key and fallback to repo key. But API keys and repo key authenticate different endpoints. I don't think one can be a fallback for another.

lib/hex/remote_converger.ex
```diff
@@ -859,31 +856,39 @@ defmodule Hex.RemoteConverger do
       # Try to get token with authentication prompting enabled
       # The OnceCache ensures only one process prompts even if multiple processes
       # detect the expired token concurrently
```
ericmj, line 858: Is this still the case? It looks like we don't use `OnceCache` in this path anymore.

AFAICT `:global.trans()` only covers credential resolution, it doesn't serialize the device flow prompting.

src/mix_hex_cli_auth.erl
```diff
@@ -0,0 +1,726 @@
...
+%% </ol>
+%%
+%% The repository name is taken from the config (`repo_name' or `repo_organization').
+%%
+%% Callbacks are taken from the `cli_auth_callbacks' key in the config map.
+%%
+%% Options:
+%% <ul>
+%% <li>`optional' - When `true' (default), proceeds without auth if none found; retries with auth on 401.</li>
+%% <li>`auth_inline' - When `true', prompts user via `should_authenticate' callback. Default is `false'.</li>
```
ericmj, line 280: Here the default is `false`, but for `with_api()` the default is `true`.

src/mix_hex_cli_auth.erl
```diff
@@ -0,0 +1,726 @@
...
+        {{?MODULE, repo, RepoName}, self()},
+        fun() ->
+            do_resolve_repo_auth(RepoName, RepoName, Config)
+        end,
+        [node()],
+        infinity
+    ).
+
+do_resolve_repo_auth(RepoName, LookupRepo, Config) ->
+    Trusted = maps:get(trusted, Config, false),
```
ericmj, line 489: If the default is `trusted: false` then we need to make sure to pass `trusted: true` on `mix hex.repo add --auth-key`.

lib/hex/update_checker.ex
```diff
@@ -82,6 +82,12 @@ defmodule Hex.UpdateChecker do
     {:noreply, state}
   end
 
+  if Mix.env() == :test do
+    def handle_info({:mix_shell, _level, _message}, state) do
+      {:noreply, state}
+    end
+  end
```
ericmj: This looks like a potential issue, why are we getting shell messages here?

lib/hex/oauth.ex
ericmj: Is this dead code?

lib/hex/repo.ex
```diff
@@ -61,18 +44,22 @@ defmodule Hex.Repo do
   end
 
   defp default_organization(repo, source, name) do
-    url = merge_values(Map.get(repo, :url), source.url <> "/repos/#{name}")
+    url = merge_values(Map.get(repo, :url), source.url)
```
ericmj: `default_organization` stores the bare base :url and sets :repo_organization.  `build_url` appends `/repos/<org>`. Existing configs with a custom org URL now double-append, `.../repos/acme/repos/acme`.

src/mix_hex_cli_auth.erl
```diff
@@ -0,0 +1,726 @@
...
+when
+    Result :: term().
+with_api(Permission, BaseConfig, Fun, Opts) ->
+    Optional = proplists:get_value(optional, Opts, false),
+    AuthInline = proplists:get_value(auth_inline, Opts, true),
+    case resolve_api_auth(Permission, BaseConfig) of
+        {ok, ApiKey, AuthContext} ->
+            Config = BaseConfig#{api_key => ApiKey},
+            execute_with_retry(Config, Fun, AuthContext, 0, undefined, Opts);
+        {error, no_auth} when Optional =:= true ->
```
ericmj, line 230: We need the optional check on `token_refresh_failed` also, like on `with_repo`: https://github.com/hexpm/hex/pull/1143/changes#diff-9618ec583c0e0fb894f3fa6fe5607b1dc95b38467b2408b4fd85ad816a423206R308

ericmj (changes requested):
Looks like we lost support for `HEX_OTP` environment variable for non-interactive use.

`Hex.OnceCache` looks like dead code now? Delete it if we no longer use it.

lib/hex/remote_converger.ex
maennchen: You are right, I only covered refresh and token exchange. I'll do a follow up PR in hex_core.

lib/hex/oauth.ex
maennchen: Yes, removed.

lib/hex/remote_converger.ex
maennchen: https://github.com/hexpm/hex_core/pull/197

lib/hex/api/oauth.ex
maennchen: Cleaned up to use atom keys consistently.

src/mix_hex_cli_auth.erl
maennchen: I did this on purpose because most repo calls did not use inline auth but most api calls did.

Would you like me tho change that?

lib/hex/auth.ex
maennchen: Yes, I missed that this branch is used for repo auth as well. Updated.

src/mix_hex_cli_auth.erl
maennchen: That's already done in `lib/mix/tasks/hex.repo.ex`

```elixir
  defp fetch_public_key(fingerprint, repo_url, auth_key, oauth_exchange) do
    repo_config = %{
      url: repo_url,
      auth_key: auth_key,
      trusted: true,
      oauth_exchange: oauth_exchange
    }
```

lib/hex/update_checker.ex
maennchen: That seems to have resolved itself. It's possible I did auth_inline somewhere where I shouldn't have when I was developing this.

Removing.

src/mix_hex_cli_auth.erl
maennchen: Handled in https://github.com/hexpm/hex_core/pull/198

src/mix_hex_cli_auth.erl
ericmj: Okay, that makes sense. Leave as is.

lib/hex/repo.ex
```diff
@@ -303,54 +297,32 @@ defmodule Hex.Repo do
...
+        [name] -> {name, :undefined}
+      end
+
     config = %{
-      :mix_hex_core.default_config()
-      | http_adapter: {Hex.HTTP, %{}},
-        repo_name: hex_to_actual_repo_name(repo_name),
+      Hex.API.Client.config()
+      | repo_name: repo_name,
+        repo_organization: organization,
```
ericmj: `build_hex_core_config` always derives `repo_organization` from the name here, ignoring the `:repo_organization` that `put_organization_url` sets. For a custom org URL, update_organizations deliberately keeps the URL verbatim and omits `:repo_organization`, but this still sets it from the name, so `build_url` appends `/repos/<org>` to the custom URL (http://custom.example/mirror -> http://custom.example/mirror/repos/acme/...). On main that URL was used as-is. For a custom URL that already contains `/repos/<org>` it double-appends. Suggest using `repo_config[:repo_organization]` (the field put_organization_url computes) and only falling back to the name when it's absent.

ericmj (approved):
Is `Hex.API.OAuth.device_authorization/2,` `poll_device_token/1`, and `refresh_token/1` dead code now?

PR #1156, opened 2026-05-13
Fix OTP archive root

### LINKED ISSUE #1154: Folder inside the archive must not contain -otp-28 prefix
Today, Hex archives have a `hex-VSN-otp-NUM.ez`. This is fine. The issue is that the folder inside the archive also has the same structure and that's invalid. Folders must have the `app-VSN` notation exclusively. This causes issues when building releases, such as: https://github.com/elixir-lang/elixir/issues/15356 - because Erlang's tooling expects `app-VSN`.

ericmj:
Closes https://github.com/hexpm/hex/issues/1154.

### DIFF
--- a/scripts/release_hex.sh
+++ b/scripts/release_hex.sh
@@ -136,14 +136,15 @@ function build {
 
   echo "Building ${elixir_version} ${otp_version} ${ubuntu_version}"
   rm -rf _build src/mix_safe_erl_term.erl
+  archive_ez=hex-${hex_version}.ez
   hex_ez=hex-${hex_version}-otp-${otp_release}.ez
 
   mkdir -p "$installs_dir/${saved_elixir_version}"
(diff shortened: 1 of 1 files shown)


maennchen:
@ericmj Is there a chance that we can patch existing releases?
For my use case specifically `2.3.1`.

I'm doing some Dependabot Core changes where this issue causes trouble. And it uses `2.3.1`. Updating (which we'll have to eventually as well) causes other problems since their test harness crashed with OAuth2 Token Exchange.

ericmj:
Yes, I can rerelease those versions.

ericmj:
Rereleased:
-   2.2.1
-   2.2.2
-   2.3.0
-   2.3.1
-   2.3.2
-   2.4.0
-   2.4.1
-   2.4.2

PR #1160, opened 2026-05-18
Add release-age cooldown to dependency resolution

ericmj:
A configurable delay between when a package version is published and when it becomes eligible for selection during dependency resolution, to mitigate supply-chain attacks where a freshly compromised release is pulled into projects before it can be detected and retired.

Configuration:
* New `cooldown` config key (env `HEX_COOLDOWN`, project `:hex` block, global `hex.config`); accepts "<N>d", "<N>w", "<N>mo".
* New `cooldown_exclude_repos` config key (env `HEX_COOLDOWN_EXCLUDE_REPOS` comma-separated, project / global as list); skips cooldown for the named repos so an organization can consume hotfixes to its own repository without delay.

Implementation:
* `Hex.Cooldown` parses durations, computes cutoffs, and renders pre-flight errors and solver failure notes.
* `Hex.Registry.Server` stores per-release `published_at` (Unix seconds extracted from the registry Timestamp) under a new ETS slot; `@ets_version` bumps to 5 so stale caches missing the slot are invalidated on upgrade. `Hex.Dev.copy_package` and the test fixture helper grow alongside.
* `Hex.Registry.Cooldown` wraps the registry behaviour, filtering versions on `versions/2` against the active cutoff and respecting the per-repo exclude list and a per-resolution bypass set.
* `Hex.RemoteConverger` builds the cutoff and bypass set at the start of resolution. The bypass set has two sources:
  - packages whose lock survived prepare_locked (mix deps.get against an intact lockfile bypasses cooldown entirely);
  - packages in old_lock whose locked version is known-unsafe (retired or carrying a security advisory) so re-resolving to escape a known-unsafe lock is never blocked. A pre-flight check walks both top-level requirements and nested hex deps under non-Hex (path / git) parents, raising a formatted cooldown error when every matching version of a direct dep is in cooldown. Solver failures get a generic cooldown hint appended when cooldown is non-zero.

Releases without `published_at` (legacy registry data, repos that have not rebuilt their index, or self-hosted clones) are treated as eligible. Empty env values (`HEX_COOLDOWN=`,
`HEX_COOLDOWN_EXCLUDE_REPOS=`) fall through to the next source instead of clobbering project / global config.

Re-vendor hex_core to pick up the Timestamp `published_at` field on `Release`.


brendon9x:
Hi @ericmj – I apologise for adding to the maintainer burden with this question, but I am wondering when you are planning on releasing this awesome change?

We're looking to rollout cooldowns across our org with some urgency and we're trying to decide whether we wait for this or mitigate it ourselves.

Thanks for the great work.

ericmj:
We want to include dependency policies in this release which is still ongoing work so we don't have a release date. If you are doing organization wide cooldowns I think it would be very relevant for your use case.

If implementing dependency policies drags on we will do an intermediate release with only cooldowns.

https://github.com/hexpm/hexpm/pull/1622

PR #1163, opened 2026-05-19
Add mix hex.search QUERY

josevalim:
It will search for the given terms and print the results to the terminal.

Package search was moved to `mix hex.package search`, as the API was getting complex (so this should likely be a minor release).


ericmj (approved):

josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #1164, opened 2026-05-19
Fix type warnings on Elixir v1.20

josevalim:


lib/hex/remote_converger.ex
```diff
@@ -613,7 +613,6 @@ defmodule Hex.RemoteConverger do
   end
 
   defp verify_deps(repo, name, version, deps)
-  defp verify_deps(nil, _name, _version, _deps), do: :ok
```
josevalim, line 616: repo is never nil because the Utils.lock sets a default value

lib/hex/registry/server.ex
```diff
@@ -495,8 +498,7 @@ defmodule Hex.Registry.Server do
         {:noreply, state}
 
       true ->
-        repo = if repo, do: "#{repo}/"
-        Mix.raise("Package #{repo}#{package} not prefetched, please report this issue")
+        Mix.raise("Package #{repo}/#{package} not prefetched, please report this issue")
```
josevalim, line 501: repo is never nil because we default to hexpm a couple lines above.

PR #1165, opened 2026-05-19
Move security advisory warning to post_converge

ericmj:
Long deps.get / deps.update output could push the "Found packages with security advisories" line past the scroll fold, hiding it. Move the summary to post_converge alongside the sponsor tip and point users back to the per-package advisory details above.


maennchen:
@ericmj Does it make sense to do the same for retirement?

PR #1166, opened 2026-05-19
update test matrix to include newer major and minor versions

Ch4s3:
I'm not sure if it's desirable to test against the latest RC of elixir like this, but I've seen other ecosystems do something like this and in any case it might be good to move the main versions up to the latest Elixir 1.19 and OTP28.5.

### DIFF
--- a/lib/hex/scm.ex
+++ b/lib/hex/scm.ex
@@ -196,6 +196,15 @@ defmodule Hex.SCM do
       raise("Checksum mismatch against registry (outer)")
     end
 
+    # Delete stale build artifact so Mix's dep loader does not read an outdated
+    # .app version and incorrectly mark the dependency as :divergedreq.  This
+    # situation arises on Elixir >= 1.20 which removed the `recently_fetched?`
+    # guard from Mix.Dep.Loader.validate_app (previously the guard prevented the
+    # loader from reading a stale .app file after a fetch).
+    if build = opts[:build] do
+      Path.join([build, "ebin", "#{name}.app"]) |> File.rm()
+    end
+
     build_tools = guess_build_tools(meta)
 
     managers =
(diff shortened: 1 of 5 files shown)


ericmj:
@Ch4s3 Looks like there is a failing test even after retrying. Can you investigate?

Ch4s3:
> @Ch4s3 Looks like there is a failing test even after retrying. Can you investigate?

I sure can. This is the exact sort of thing I thought running against the latest versions might uncover.

Ch4s3:
Interesting, this is a fixture issue. The data is wrong and new versions of mix now catch it. I'll have a fix later.

lib/hex/scm.ex
```diff
@@ -196,6 +196,15 @@ defmodule Hex.SCM do
...
     end
 
+    # Delete stale build artifact so Mix's dep loader does not read an outdated
+    # .app version and incorrectly mark the dependency as :divergedreq.  This
+    # situation arises on Elixir >= 1.20 which removed the `recently_fetched?`
+    # guard from Mix.Dep.Loader.validate_app (previously the guard prevented the
+    # loader from reading a stale .app file after a fetch).
+    if build = opts[:build] do
+      Path.join([build, "ebin", "#{name}.app"]) |> File.rm()
+    end
```
Ch4s3: @ericmj you probably understand mix way better than I do. Can you think of an approach that doesn't require a call to  `File.rm/0`?

ericmj: This looks like an Elixir bug to me. Opened a PR with a proposed fix here: https://github.com/elixir-lang/elixir/pull/15406

Ch4s3: That's even better. I can point this to the next rc, main, or the eventual 1.20 when that's ready.

PR #1168, opened 2026-05-22
Dependency policies

ericmj:
Adds client-side enforcement of dependency policies. A policy is defined on hex.pm under an organization (or on any self-hosted repo) and fetched through the registry at resolution time. The client filters candidate versions the policy blocks before the solver sees them, so resolution simply never picks them.

Per repository, a policy can express:

* a cooldown (minimum age before a newly published version becomes eligible)
* an advisory rule (block versions with security advisories at or above a severity threshold, or with any advisory)
* a retirement rule (block versions retired for the configured reasons)
* package/version overrides (allow or deny exceptions; the most specific match wins, and an allow exempts the release from the restriction entirely)

Versions already in the lockfile are exempt from filtering, so re-resolution keeps a locked-but-now-blocked entry instead of failing. Policies have a visibility: public policies are fetchable anonymously, private ones require authentication to the owning organization.

## Configuring the policy

A project has exactly one active policy, resolved with the usual Hex config precedence: env var, then `mix.exs`, then global config.

`HEX_POLICY` env var, a `REPO/NAME` pair. Overrides everything for the invocation, and an empty value disables the configured policy:

```
$ HEX_POLICY=hexpm:myorg/strict-prod mix deps.get
$ HEX_POLICY= mix deps.get
```

`mix.exs` `:hex` block, with `org:` for a hexpm organization or `repo:` for any configured repo:

```elixir
defp project() do
  [hex: [policy: [org: "myorg", name: "strict-prod"]]]
end
```

`mix hex.config`, stored in the global Hex config:

```
$ mix hex.config policy hexpm:myorg/strict-prod
```

The bare `hexpm` repo is rejected in all forms: the global repository has no organization-scoped policies, so policies live under `hexpm:<org>` or a self-hosted repo.

## New mix tasks

`mix hex.policy show` summarizes the active policy: its visibility, the per-repository restrictions (cooldown, advisory rule, retirement rule) and override counts, plus the effective cooldown, which is the strictest of the local `cooldown` config and the policy's own.

`mix hex.policy why PACKAGE` (or `REPO/PACKAGE`) walks every version of the package in the registry, classifies each against the active policy, and prints a per-version table:

```
Versions of "plug" (47):

Version  Status   Blocked by
1.16.0   ALLOWED
1.17.0   BLOCKED  retirement: security
1.18.0   BLOCKED  advisory ≥ high, cooldown 14d; eligible 2026-06-20
```

This is the uncapped companion to the resolution output below. `mix hex.config policy` reads and writes the setting like any other config key.

## How policies surface in mix deps.get

After a successful resolution, a summary block reports the active policy, the cooldown it imposes (a locally configured cooldown keeps its own existing report), and every candidate version it hid, capped at the 5 newest per package:

```
Active policy: hexpm:myorg/strict-prod
Effective cooldown: 14d (hexpm:myorg/strict-prod)
Policy hid 7 candidate versions:
  phoenix 1.8.1 — cooldown 14d; eligible 2026-06-18
  plug 1.18.0 — advisory ≥ high
  ...and 5 more — run `mix hex.policy why plug`
```

When resolution fails and the policy hid versions of the packages involved, the solver's failure message is followed by a note attributing the hidden versions, in the same capped format, so a "no compatible versions" error explains itself:

```
Note: active policy hides 3 versions of "plug":
  plug 1.18.0 — advisory ≥ high
  ...
```

Because a policy is an enforcement feature, anything short of materializing it fails closed rather than resolving unenforced:

* a malformed config value aborts with `Invalid policy configuration: ...` and the expected formats
* a failed fetch prints `Failed to fetch policy REPO/NAME from registry` and aborts, unless a previously fetched copy is cached, in which case the cache is used and the message says so
* on a 404 or 401 the error explains the policy may not exist, may be misspelled, or may be private without sufficient permissions, and points at `mix hex.user auth` when no authenticated user is present


lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+  @severity_order [
+    :SEVERITY_NONE,
+    :SEVERITY_LOW,
+    :SEVERITY_MEDIUM,
+    :SEVERITY_HIGH,
+    :SEVERITY_CRITICAL
+  ]
+
+  @type policy :: map()
+  @type candidate :: map()
```
maennchen, line 16:
```suggestion
  @type policy :: :hex_pb_policy.Policy()
  @type candidate :: %{
    repo: String.t(),
    package: String.t(),
    version: String.t(),
    advisories: [:hex_pb_policy.SecurityAdvisory()],
    retired: :hex_pb_policy.RetirementStatus(),
    published_at: DateTime.t()
  }
```

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
+defmodule Hex.Policy.Filter do
```
maennchen, line 1: Should this be a helper module in `hex_core`?

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+  end
+
+  # Finds the RepositoryPolicy whose repository matches the candidate's repo.
+  # Candidate repos are named "hexpm" or "hexpm:<org>"; the policy's tabs use
+  # "hexpm" or the bare org name, so the "hexpm:" prefix is stripped to match.
+  defp repository_policy(policy, candidate) do
+    repo = candidate_repo(candidate)
+    key = strip_prefix(repo)
+
+    Enum.find(Map.get(policy, :repositories, []), fn rp ->
```
maennchen, line 74: Can this field be absent / not a list?


```suggestion
    Enum.find(policy.repositories, fn rp ->
```

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+      retired: Server.retired(repo, package, version_str),
+      published_at: Server.published_at(repo, package, version_str)
+    }
+  end
+
+  # Finds the RepositoryPolicy whose repository matches the candidate's repo.
+  # Candidate repos are named "hexpm" or "hexpm:<org>"; the policy's tabs use
+  # "hexpm" or the bare org name, so the "hexpm:" prefix is stripped to match.
+  defp repository_policy(policy, candidate) do
+    repo = candidate_repo(candidate)
```
maennchen, line 71: You are setting a default above.

```suggestion
    repo = candidate.repo
```

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+
+  defp strip_prefix("hexpm:" <> org), do: org
+  defp strip_prefix(repo), do: repo
+
+  # Returns the effect of the most specific override matching the candidate:
+  # :allow, :deny, or :none. A requirement-bearing override is more specific
+  # than a bare-package one for the same package.
+  defp override(repo_policy, candidate) do
+    repo_policy
+    |> Map.get(:overrides, [])
```
maennchen, line 90:
```suggestion
    repo_policy.overrides
```

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+  defp known_action?(%{action: action}),
+    do: action in [:OVERRIDE_ACTION_ALLOW, :OVERRIDE_ACTION_DENY]
+
+  defp override_matches?(%{ref: %{package: package} = ref}, %{package: package} = candidate) do
+    case Map.get(ref, :requirement) do
+      nil -> true
+      "" -> true
+      requirement -> version_satisfies?(candidate.version, requirement)
+    end
+  end
```
maennchen, line 111:
```suggestion
  defp override_matches?(%{ref: %{package: package, requirement: nil} = ref}, %{package: package} = candidate), do: true
  defp override_matches?(%{ref: %{package: package, requirement: ""} = ref}, %{package: package} = candidate), do: true
  defp override_matches?(%{ref: %{package: package, requirement: requirement} = ref}, %{package: package} = candidate), do: version_satisfies?(candidate.version, requirement)
```

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+
+  defp override_matches?(_override, _candidate), do: false
+
+  defp override_specificity(%{ref: ref}) do
+    case Map.get(ref, :requirement) do
+      nil -> 0
+      "" -> 0
+      _ -> 1
+    end
+  end
```
maennchen, line 121:
```suggestion
  defp override_specificity(%{ref:%{requirement: nil}}), do: 0
  defp override_specificity(%{ref:%{requirement: ""}}), do: 0
  defp override_specificity(%{ref:%{requirement: _}}), do: 1
```

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+      restriction ->
+        reasons =
+          []
+          |> add_advisory(restriction, candidate)
+          |> add_retirement(restriction, candidate)
+          |> add_cooldown(restriction, candidate, opts)
+
+        if reasons == [], do: :allowed, else: {:blocked, reasons}
+    end
+  end
```
maennchen, line 140:
```suggestion
  defp restriction(%{restriction: nil}, _candidate, _opts), do: :allowed
  defp restriction(%{restriction: restriction} = repo_policy, candidate, opts) do
    reasons =
        []
        |> add_advisory(restriction, candidate)
        |> add_retirement(restriction, candidate)
        |> add_cooldown(restriction, candidate, opts)

      if reasons == [], do: :allowed, else: {:blocked, reasons}
  end
```

lib/hex/policy.ex
```diff
@@ -0,0 +1,126 @@
...
+    * a keyword list in mix.exs: `[org: "myorg", name: "strict-prod"]` for a
+      hexpm organization (resolves to the `hexpm:myorg` repo), or
+      `[repo: "REPO", name: "NAME"]` for any configured repo
+    * a `"REPO/NAME"` string (env-var / `mix hex.config` form), e.g.
+      `"hexpm:myorg/strict-prod"`
+    * `nil` or `""` (no policy)
+
+  Returns `{:ok, string}`, `{:ok, nil}`, or `:error`. The bare `"hexpm"` repo
+  is rejected because the global hexpm has no organization-scoped policies;
+  policies live under `hexpm:<org>` (or any non-`hexpm` repo for self-hosted
```
maennchen, line 22: Should we document that restriction? We might add a `recommended` policy or something like that in the future.

lib/hex/repo.ex
```diff
@@ -228,6 +229,43 @@ defmodule Hex.Repo do
...
+
+    case put_repo_organization(config, repo_config) do
+      {:ok, config} -> :mix_hex_repo.get_policy(config, name)
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  # Policies live one level above the organization repo, so the base URL is
+  # derived by stripping the `/repos/<org>` suffix. A custom :url that doesn't
+  # follow that layout gives no base to derive the policy URL from.
```
maennchen: Why do we use that layout?

lib/hex/policy/filter.ex
ericmj: Yes, we can move there it in the future.

lib/hex/policy/filter.ex
```diff
@@ -0,0 +1,210 @@
...
+    key = strip_prefix(repo)
+
+    Enum.find(Map.get(policy, :repositories, []), fn rp ->
+      rp.repository == repo or rp.repository == key
+    end)
+  end
+
+  defp candidate_repo(%{repo: repo}) when is_binary(repo), do: repo
+  defp candidate_repo(_), do: "hexpm"
+
```
ericmj, line 81:
```suggestion
```

lib/hex/repo.ex
ericmj: The comment is incorrect, policies live under `/repos/<repo>`, another function is expecting the base URL. I will try to clean up the code so we don't have to strip.

maennchen (approved):

maennchen (approved):

PR #1171, opened 2026-05-27
Annotate cooldown-held versions in mix hex.outdated

### LINKED ISSUE #1113: Add a cooldown option to `hex.outdated`
## What

It would be nice to have a flag that filters out outdated packages based on their date of publication.

For example, `mix hex.outdated --cooldown=7-days` will mark outdated packages that were published less than 7 days ago as `Update not possible` despite their versions.

## Why

The motivation behind it is that some supply chain attacks could be mitigated or softened by introduction of dependency cooldowns: https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

Another thing is that even minor tweaks create unexpected consequences. For example [Default response_encodings change in 1.10.0 breaks HTTP client compatibility · Issue #557 · mtrudel/bandit](https://github.com/mtrudel/bandit/issues/557) was introduced partly by changing compression priority (there was a larger bug with compression going on underneath, but it was hard to notice due to the compression algorithm in question being rarely if ever called).

All in all, taking a more conservative stance on dependency updates could be beneficial. On the other hand, I'm not sure whether this approach is something that the Hex team wants to "promote" by adding cooldown flags.

## Next

It would be nice to hear thoughts from the Hex team regarding this issue. Whether this approach makes sense in general and whether it is applicable to the `hex.outdated` command.

If it does sound like something worth adding to `hex.outdated`, I am more than willing to help.

ericmj:
Show "(cooldown)" alongside the status when the latest version of a dependency is within the configured cooldown window, with a "Versions in cooldown" legend below the table listing eligibility dates. The single-package output gets a similar line, and the JSON output gains a `cooldown` field exposing published_at and eligible_on.

Closes https://github.com/hexpm/hex/issues/1113.

### DIFF
--- a/lib/hex/remote_converger.ex
+++ b/lib/hex/remote_converger.ex
@@ -143,7 +143,7 @@ defmodule Hex.RemoteConverger do
   def build_cooldown_bypass(_old_lock, _locked, :disabled), do: MapSet.new()
 
   def build_cooldown_bypass(old_lock, locked, _cutoff) do
-    # The bypass set has two sources:
+    # The bypass set has three sources:
     #
     # 1. Packages in `locked` (the post-`prepare_locked/3` set): the lockfile
     #    entry survived requirement checking and dep unlocking, so this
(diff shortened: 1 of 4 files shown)


maennchen (approved):
Looking great :)

Image
Image

maennchen (approved):

ericmj:
Thanks for the reviews @maennchen!

PR #1180, opened 2026-06-16
Fix crash when a server sends an x-hex-message header

ericmj:
`Hex.HTTP.handle_hex_message/1` is meant to surface `x-hex-message` headers from the server (e.g. deprecation notices) as `API warning:` / `API error:` output. It runs `:binary.list_to_bin/1` on the header value, which requires a charlist — but `decode_header/1` converts every response header value to a binary via `List.to_string/1`. So any real `x-hex-message` header raises `ArgumentError: not an iodata term` inside the HTTP task and fails the request (e.g. `mix deps.get`).

No hex.pm endpoint currently sends `x-hex-message`, so the bug has been latent — the existing test only exercised the function with charlists. This switches to `IO.iodata_to_binary/1`, which accepts both binaries (the real path) and charlists (the existing test path), and adds a binary regression case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Bug fixes
 
 * Fix a crash when a server responds with an `x-hex-message` header
+* Escape terminal control sequences in server-provided `x-hex-message` headers
 
 ## v2.4.2 (2026-04-30)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 * Deprecate authenticating to organization repositories with a stored key; a future release will require `mix hex.user auth` or a short-lived organization token
 * Deprecate `HEX_REPOS_KEY`; authenticate per organization with `mix hex.organization auth ORGANIZATION --key KEY`
 
+### Bug fixes
+
+* Fix a crash when a server responds with an `x-hex-message` header
+
 ## v2.4.2 (2026-04-30)
 
 ### Enhancements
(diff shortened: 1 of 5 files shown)


maennchen:
I think this bug was a blessing in disguise. This would also be called for untrusted servers, right?

There's quite a lot of terminal escape sequences that could potentially be problematic. We should add some escaping.

maennchen:
Something along those lines:

```elixir
defmodule TerminalSafe do
  @bidi_controls [
    0x061C,
    0x202A,
    0x202B,
    0x202C,
    0x202D,
    0x202E,
    0x2066,
    0x2067,
    0x2068,
    0x2069
  ]

  def escape(binary) when is_binary(binary) do
    binary
    |> String.to_charlist()
    |> Enum.map_join(fn
      cp when cp in @bidi_controls ->
        "\\u{" <> Integer.to_string(cp, 16) <> "}"

      ?\n ->
        "\n"

      ?\t ->
        "\t"

      # Important terminal controls
      ?\e ->
        "\\e"

      ?\r ->
        "\\r"

      ?\b ->
        "\\b"

      ?\a ->
        "\\a"

      # Other C0 controls
      cp when cp < 0x20 ->
        hex =
          cp
          |> Integer.to_string(16)
          |> String.upcase()
          |> String.pad_leading(2, "0")

        "\\x" <> hex

      # DEL
      0x7F ->
        "\\x7F"

      # C1 controls
      cp when cp >= 0x80 and cp <= 0x9F ->
        "\\u{" <> Integer.to_string(cp, 16) <> "}"

      # Normal printable codepoint
      cp ->
        <<cp::utf8>>
    end)
  end
end
```

maennchen (approved):

PR #1184, opened 2026-06-17
Clear the OAuth token and warn when its refresh fails

ericmj:
An expired global OAuth token whose refresh the server rejects was refreshed once per package during dependency resolution -- dozens of failed POSTs to /oauth/token, serialized by the token-refresh lock -- and the failure was swallowed, so resolution silently continued without credentials. This regressed the OnceCache memoization and warning dropped in cda8717.

Implement the new hex_cli_auth clear_oauth_tokens callback: drop the token from in-memory state (keeping the on-disk token so a transient failure can recover on the next run) and warn once that the session expired and to re-run `mix hex.user auth`. With the token gone, the rest of the run stops retrying the refresh.

Re-vendors hex_core.

### DIFF
--- a/lib/hex/auth.ex
+++ b/lib/hex/auth.ex
@@ -40,6 +40,7 @@ defmodule Hex.Auth do
       get_auth_config: &get_auth_config/1,
       get_oauth_tokens: &get_oauth_tokens/0,
       persist_oauth_tokens: &persist_oauth_tokens/4,
+      clear_oauth_tokens: &clear_oauth_tokens/0,
       prompt_otp: &prompt_otp/1,
       get_client_id: &Hex.API.OAuth.client_id/0,
       should_authenticate: &should_authenticate/1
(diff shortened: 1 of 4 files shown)


lib/hex/auth.ex
```diff
@@ -111,6 +112,22 @@ defmodule Hex.Auth do
     :ok
   end
 
+  # Invoked by hex_cli_auth when the stored global OAuth token is expired and
+  # could not be refreshed. Drop it from in-memory state only — keeping the
+  # on-disk token, since a refresh can fail transiently — so the rest of this
+  # run stops retrying the doomed refresh and proceeds unauthenticated.
+  defp clear_oauth_tokens do
+    Hex.State.put(:oauth_token, nil)
```
maennchen:
```suggestion
    Hex.OAuth.clear_tokens()
```

maennchen:
Oh wait, you wanted this behaviour: `test clear_oauth_tokens/0 callback keeps the on-disk token so a later run can retry the refresh (Hex.AuthTest)`

So your code was correct.

maennchen:
But the question then is if it makes sense. 500 etc. should not call clear and 401 etc. will never work in the future.

ericmj:
We want to retry on 500 and a 401 will trigger a new device auth, like `mix hex.user auth` so this is behavior we want I think

maennchen (approved):
True. The whole auth logic has so many paths 😅

PR #1193, opened 2026-06-29
feat: list all package overrides in mix hex.policy show

### LINKED ISSUE #1190: `mix hex.policy show` should list all package overrides
Currently it only shows the count:

```
$ mix hex.policy show
Active policy: hex/production [public]

    hexpm:
      Cooldown:        14d
      Advisory rule:   block ≥ MEDIUM
      Retirement rule: SECURITY, DEPRECATED, RENAMED
      Overrides:       2

Effective cooldown: 14d (hex/production)
```

We should also remove the implicit `show` when running `mix hex.policy`. Running just `mix hex.policy` should be an error.

mvanhorn:
## Summary
`mix hex.policy show` now lists each package override per repository (package name, version requirement when present, and ALLOW or DENY) instead of printing only a count, so users can see exactly which packages a policy explicitly allows or denies. A repository with no overrides renders a `(none)` placeholder.

## Why this matters
Issue #1190 asks for the overrides to be enumerated rather than summarized as a number. The old `render_repository/1` in `lib/mix/tasks/hex.policy.ex` printed `Overrides: <count>`, which gave no insight into which packages were affected. This change also removes the implicit `show` default: a bare `mix hex.policy` previously ran `show`, and now falls through to the existing `Mix.raise` usage message so the invocation is explicit.

## Testing
`mix test test/mix/tasks/hex.policy_test.exs` (8 passed). Added `show` cases asserting each override line renders the package, requirement, and ALLOW/DENY, and that an empty override list prints `(none)`. Replaced the old "bare invocation defaults to show" test with one asserting `run([])` raises the usage `Mix.Error`.

Fixes #1190

### DIFF
--- a/lib/mix/tasks/hex.policy.ex
+++ b/lib/mix/tasks/hex.policy.ex
@@ -9,7 +9,6 @@ defmodule Mix.Tasks.Hex.Policy do
   Shows the active Hex policy and explains why specific versions
   are blocked.
 
-      $ mix hex.policy
       $ mix hex.policy show
       $ mix hex.policy why PACKAGE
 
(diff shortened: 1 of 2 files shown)


ericmj:
Thank you! 💜

PR #1202, opened 2026-07-17
Print warnings to stderr instead of stdout

maennchen:
`Hex.Shell.warn/1` wrote to stdout via `Mix.shell().info/1`, so warnings like the expired-authentication notice or the stale mix.lock version warning polluted machine-readable stdout output such as `mix hex.outdated --json`. Route `warn/1` through `Mix.shell().error/1`, which prints to stderr, keeping the yellow warning color.

Warning-colored lines that are part of a command's regular stdout output (the retirement reason in the `mix deps.get` dependency listing and the retirement notice in `mix hex.info`) stay on stdout and are now printed as yellow-formatted info lines instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v2.5.2-dev
 
+### Enhancements
+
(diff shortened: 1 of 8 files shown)


ericmj:
Thank you! 💜

PR #1203, opened 2026-07-17
Add SARIF output format to mix hex.audit

maennchen:
Add a `--format` sarif option that renders the audit result as a SARIF v2.1.0 document, plus an `--output PATH` option to write it to a file instead of stdout. This allows uploading audit findings to GitHub code scanning and other SARIF consumers; the`moduledoc` documents a complete GitHub Actions workflow using `github/codeql-action/upload-sarif`.

Retirements map to one rule per retirement reason (HEX0001 to HEX0005, with security retirements defaulting to level error) and advisories map to one rule per advisory identifier so each alert carries its own severity score, title and help link. Results are anchored to the dependency entry in `mix.lock`, relative to the git repository root when available, with region and context region snippets. The run includes git-derived version control provenance (with credentials stripped from
the repository URL), stable partial fingerprints, and messages that carry both resolved text and template arguments. Ignored findings are included with a suppression so they show up as closed alerts.

The exit code behaves the same as with the human format. SARIF output requires the OTP 27 `json` module, matching `mix hex.outdated --json`.

Needs #1202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 * Print warnings to standard error instead of standard output, keeping stdout clean for machine-readable output such as `mix hex.outdated --json`. Warning-colored lines that are part of a command's regular output, such as retirement notices in `mix hex.info` and the `mix deps.get` dependency listing, remain on standard output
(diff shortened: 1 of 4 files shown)


ericmj:
Thank you! 💜

PR #1217, opened 2026-08-04
Ask about SSO only where the answer would change anything

ericmj:
Prompts once, at the terminal, when an organization the project depends on needs its SSO session renewed.

`mix deps.get` already knows every repository it needs before it fetches anything, because a published package's dependencies can only come from the public repository or the package's own organization. That set is intersected with the organizations the last token refresh flagged, so a member of ten SSO organizations who depends on two is asked about two:

```
acme requires SSO authentication. Authenticate now? [Yn]
```

On yes it requests a re-authorization URI, prints it, and waits. Completing it in a browser renews the existing session rather than replacing it, so the refresh token survives and nobody re-runs device auth. On no it warns and carries on without those packages.

It asks only where the answer would change something. A repository authenticated with an organization key never touches the stored token, so nothing about it is worth asking. Offline says so instead of prompting, and `HEX_API_KEY` says it authenticates as itself.

Vendors hexpm/hex_core#213, which has to merge first.


lib/hex/remote_converger.ex
```diff
@@ -1012,11 +1012,20 @@ defmodule Hex.RemoteConverger do
...
         finish_sso_reauth(organizations)
 
+      # The server's message never names a client command, since it cannot know
+      # which client asked, so the mix task goes in here. A full `mix hex.user
+      # auth` re-establishes organization access at approval, which makes it
+      # the fallback whatever kept the in-place flow from starting.
       {:ok, {_status, _headers, %{"message" => message}}} when is_binary(message) ->
-        Hex.Shell.warn("Could not start SSO authentication: #{message}")
+        Hex.Shell.warn(
+          "Could not start SSO authentication: #{message}. " <>
```
maennchen: We should probably send this through `Hex.Util.escape_terminal` to be safe.

maennchen (approved):
Looks good to me besides the nit.

PR #1237, opened 2026-09-09
Verify organization 2FA and SSO requirements before fetching

ericmj:
When private dependencies need organization authentication, Hex names the affected organizations and displays separate 2FA enrollment and SSO requirements. The shared browser flow completes outstanding requirements before the client refreshes scopes and resumes fetching, and outstanding requirements are kept when the flow is cancelled, expires, or receives EOF.

Vendor the corresponding hex_core protocol changes and add integration coverage against Hexpm for organization scope filtering, verification requests, and an enrolled member keeping access when a policy starts. The integration harness starts the server with organization 2FA policies enabled. The full Hex suite against the changed server passed 662 tests with one existing skip; formatting passed.

Vendors https://github.com/hexpm/hex_core/pull/222, merged as https://github.com/hexpm/hex_core/commit/7710855; the vendored files match that commit.

Related: https://github.com/hexpm/hexpm/pull/1913, https://github.com/hexpm/hex_core/pull/222.


.github/workflows/main.yml
```diff
@@ -68,7 +68,7 @@ jobs:
     env:
       HEXPM_OTP: OTP-28.5
       HEXPM_ELIXIR: v1.19.5
-      HEXPM_BRANCH: main
+      HEXPM_BRANCH: organization-tfa-enforcement
```
ericmj, line 71: This needs reverting back to `main` before merging.

